PR #304, opened 2023-07-29
feat: add order notification schema to README

rileydcampbell:
## Summary

Add order notification webhook schema.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,3 +62,21 @@ PERMIT_TENDERLY=<>
 1. If failures, look at logs in Cloudwatch Insights
 
 1. Repeat
+
+## Order Notification Schema
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -62,3 +62,21 @@ PERMIT_TENDERLY=<>
 1. If failures, look at logs in Cloudwatch Insights
 
 1. Repeat
+
+## Order Notification Schema
+
+Depending on the filler preferences, the notification webhook can POST orders with a specific exclusive filler address or all new orders. The following schema is what the filler execution endpoint can expect to receive.
```
zhongeric, line 68: Perhaps we should use the term `quoters` here since I think we made the distinction before in the docs between the two. Like normal fillers won't have access to this feature unless enrolled in rfq.

zhongeric (approved):

README.md
marktoda: I could imagine this feature being useful (ie not for exclusive filler query but maybe some other more generic query) in future for public fillers as well

PR #305, opened 2023-07-29
feat: update swagger docs

rileydcampbell:


swagger.json
```diff
@@ -91,14 +94,14 @@
...
       "get": {
-        "tags": ["Dutch Auction"],
+        "tags": ["Dutch Order"],
         "summary": "Get nonce",
         "description": "Get current nonce for dutch auction orders. Given an address this endpoint will return the next valid nonce to be used in order creation.",
         "parameters": [
           {
-            "$ref": "#/components/parameters/limitParam"
+            "$ref": "#/components/parameters/addressParam"
           }
```
zhongeric, line 105: note that nonce also takes an optional chainId argument

swagger.json
```diff
@@ -422,6 +430,7 @@
       },
       "PostOrderRequest": {
         "type": "object",
+        "required": ["encodedOrder", "signature", "chainId"],
```
zhongeric, line 433: should we add `quoteId` here too? they would get this from URA and it does help a lot in debugging

zhongeric (approved):

marktoda (approved):

swagger.json
rileydcampbell: Hmmm we could but it's not actually required

swagger.json
rileydcampbell: Added chainId here and forgot to also include on the GET /orders parameters

PR #311, opened 2023-08-09
fix: checksum filler address in post-order log

ConjunctiveNormalForm:
archivedOrders table store checksummed filler address so we should do the same for postedOrders otherwise SQL string equality comparison would fail


rileydcampbell (approved):
Lgtm, I'd just be worried about breaking some of the data team dashboards. Should be okay though?

PR #316, opened 2023-09-06
build: upgrade to node v18

ConjunctiveNormalForm:


rileydcampbell (approved):

bin/app.ts
rileydcampbell, line 1: Need to upgrade the build image too I think
https://docs.aws.amazon.com/codebuild/latest/userguide/available-runtimes.html

PR #328, opened 2023-10-03
feat: add gs reaper cron job

ConjunctiveNormalForm:
Completes [[GS] Reaper job](https://github.com/orgs/Uniswap/projects/22/views/28?pane=issue&itemId=34655244)

Query and deletes all open orders for which deadline is more than an hour in the past

Also tested in my dev account

### DIFF
--- a/bin/stacks/cron-stack.ts
+++ b/bin/stacks/cron-stack.ts
@@ -1,0 +1,43 @@
+import * as cdk from 'aws-cdk-lib'
+import * as aws_events from 'aws-cdk-lib/aws-events'
+import * as aws_iam from 'aws-cdk-lib/aws-iam'
+import * as aws_lambda_nodejs from 'aws-cdk-lib/aws-lambda-nodejs'
+import { Construct } from 'constructs'
+import path from 'path'
+
+import { SERVICE_NAME } from '../constants'
+
+export interface CronStackProps extends cdk.NestedStackProps {
+  lambdaRole: aws_iam.Role
+  chatbotSNSArn?: string
+  envVars?: { [key: string]: string }
+}
+
+export class CronStack extends cdk.NestedStack {
+  public readonly gsReaperCronLambda?: aws_lambda_nodejs.NodejsFunction
+
+  constructor(scope: Construct, name: string, props: CronStackProps) {
+    super(scope, name, props)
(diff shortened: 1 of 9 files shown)


bin/stacks/cron-stack.ts
rileydcampbell, line 1: nit: change name of file/stack to `reaper-cron-stack`, think in other repos we've started using application specific stack naming

lib/crons/gs-reaper.ts
```diff
@@ -0,0 +1,47 @@
...
+        return order.orderHash
+      }
+      return []
+    })
+    log.info({ staleOrders }, `Found ${staleOrders.length} stale orders`)
+    if (staleOrders.length > 0) {
+      try {
+        await repo.deleteOrders(staleOrders)
+      } catch (e) {
+        log.error({ error: e }, 'Failed to delete stale orders')
```
rileydcampbell, line 37: Maybe add a metric here so we can add alarms based on cron failures.

rileydcampbell (approved):
Couple nits but otherwise lgtm!

bin/stacks/cron-stack.ts
ConjunctiveNormalForm: really? so we'd have one cron stack for each cron job? not sure if I agree with the approach tbh 😅

marktoda (approved):
lgtm

bin/stacks/cron-stack.ts
rileydcampbell: eh yeah maybe not then

rileydcampbell: I'm good with this

bin/stacks/cron-stack.ts
```diff
@@ -0,0 +1,43 @@
...
+
+import { SERVICE_NAME } from '../constants'
+
+export interface CronStackProps extends cdk.NestedStackProps {
+  lambdaRole: aws_iam.Role
+  chatbotSNSArn?: string
+  envVars?: { [key: string]: string }
+}
+
+export class CronStack extends cdk.NestedStack {
```
rileydcampbell, line 16: Doesn't have to be in this PR but would add the alarms here for the `DeleteStaleOrdersError` metric.

rileydcampbell (approved):

PR #329, opened 2023-10-18
fix: posted order log

ConjunctiveNormalForm:
Only log user output (defined by largest of outputs, assuming same token)

### DIFF
--- a/lib/entities/Order.ts
+++ b/lib/entities/Order.ts
@@ -17,9 +17,9 @@ export type OrderInput = {
 
 export type OrderOutput = {
   token: string
-  startAmount?: string
-  endAmount?: string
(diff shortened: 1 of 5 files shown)


lib/entities/Order.ts
```diff
@@ -17,9 +17,9 @@ export type OrderInput = {
 
 export type OrderOutput = {
   token: string
-  startAmount?: string
-  endAmount?: string
-  recipient?: string
+  startAmount: string
+  endAmount: string
+  recipient: string
```
ConjunctiveNormalForm, line 22: these should never be undefined afaict?

marktoda (approved):

PR #330, opened 2023-10-20
fix: only log user settled amounts

ConjunctiveNormalForm:
Right now for each fill with two outputs (user amount & fee), we log two events. We don't need the fee one since we have a completely separate source of truth for fees

### DIFF
--- a/lib/entities/Order.ts
+++ b/lib/entities/Order.ts
@@ -23,10 +23,10 @@ export type OrderOutput = {
 }
 
 export type SettledAmount = {
-  tokenOut?: string
-  amountOut?: string
-  tokenIn?: string
-  amountIn?: string
+  tokenOut: string
+  amountOut: string
+  tokenIn: string
+  amountIn: string
 }
 
 export type OrderEntity = {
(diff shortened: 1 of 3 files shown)


lib/entities/Order.ts
```diff
@@ -23,10 +23,10 @@ export type OrderOutput = {
...
 
 export type SettledAmount = {
-  tokenOut?: string
-  amountOut?: string
-  tokenIn?: string
-  amountIn?: string
+  tokenOut: string
+  amountOut: string
+  tokenIn: string
+  amountIn: string
```
ConjunctiveNormalForm, line 29: those should always be defined no?

test/handlers/check-order-status.test.ts
```diff
@@ -154,6 +159,58 @@ describe('Testing check order status handler', () => {
       })
     })
 
+    it('should check fill events when order expired', async () => {
```
ConjunctiveNormalForm, line 162: unit tests for uncovered branches unrelated to this PR :p to pass jest coverage check

PR #334, opened 2023-11-06
chore: remove provisioned db capacity

rileydcampbell:
## Summary
Remove provisioned capacity for DynamoDB. We should switch to the auto scaling and PAY_PER_REQUEST set up. Our usage is so low I don't think it make sense to provision this. All of the usage charts for indexes look like the below chart.
Screenshot 2023-11-06 at 2 46 30 PM

### DIFF
--- a/bin/app.ts
+++ b/bin/app.ts
@@ -11,7 +11,7 @@ import dotenv from 'dotenv'
 import 'source-map-support/register'
 import { SUPPORTED_CHAINS } from '../lib/util/chain'
 import { STAGE } from '../lib/util/stage'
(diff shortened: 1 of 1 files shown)


zhongeric (approved):
Lgtm but this seems like it just removes the provisioning on the indexes and not the table overall?

rileydcampbell:
> Lgtm but this seems like it just removes the provisioning on the indexes and not the table overall?

Yep, you have to do one at a time. It's index provisioned, table provisioning, then the table billing mode. Lol just some CDK changeset rules i think.

PR #341, opened 2023-12-19
chore: add speedy jest compilation

rileydcampbell:
## Summary
Saw the backend team started using speedy web compiler and tested it out for UniswapX service. Works well!

ts-jest has a 6 - 12 second lag when compiling. This compilation lag gets extra bad if you run jest in watch mode. Instead, we can use [swc](https://swc.rs/docs/usage/jest#usage) as a drop-in replacement to reduce the lag to one second.

### Before SWC
Screenshot 2023-12-19 at 11 01 39 AM

### After SWC
Screenshot 2023-12-19 at 11 01 01 AM

### DIFF
--- a/test/handlers/post-order.test.ts
+++ b/test/handlers/post-order.test.ts
@@ -1,7 +1,7 @@
-const parserMock = jest.fn()
-
+/* eslint-disable @typescript-eslint/ban-ts-comment */
+// @ts-nocheck
 import { SFNClient, StartExecutionCommand } from '@aws-sdk/client-sfn'
-import { OrderType, OrderValidation, REACTOR_ADDRESS_MAPPING } from '@uniswap/uniswapx-sdk'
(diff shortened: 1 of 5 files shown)


test/handlers/post-order.test.ts
```diff
@@ -45,9 +45,7 @@ jest.mock('@uniswap/uniswapx-sdk', () => {
   const originalSdk = jest.requireActual('@uniswap/uniswapx-sdk')
   return {
     ...originalSdk,
-    DutchOrder: {
-      parse: parserMock,
```
ConjunctiveNormalForm, line 49: v nit but why do do we need to do this?

ConjunctiveNormalForm (approved):

test/handlers/post-order.test.ts
rileydcampbell: There are some differences in how SWC compiles our code. It doesn't compile to Babel. Something about the way jest mocking uses hosting that isn't compatible with SWC. Here's a couple issues on it if you're curious.
https://github.com/jestjs/jest/issues/9430
https://github.com/swc-project/swc/discussions/7024

PR #344, opened 2024-01-19
feat: add glob to webhook definition

zzmp:
Adds a glob to the webhook definition, to allow the JSON WebhookDefinition to include a catch-all that is sent every order and order update.


zzmp:
Closing in favor of #345 (which is identical, but from the same remove).

PR #345, opened 2024-01-22
feat: add glob to webhook definition

zzmp:
Adds a glob to the webhook definition, to allow the JSON WebhookDefinition to include a catch-all that is sent every order and order update.

### DIFF
--- a/lib/providers/json-webhook-provider.ts
+++ b/lib/providers/json-webhook-provider.ts
@@ -15,16 +15,18 @@ export class JsonWebhookProvider implements WebhookProvider {
 }
 
 export function findEndpointsMatchingFilter(filter: OrderFilter, definition: WebhookDefinition): Webhook[] {
-  let endpoints: Webhook[] = definition['*'] ?? []
+  const endpoints: Webhook[] = []
+
+  const catchallEndpoints = definition['*'] ?? []
+  endpoints.push(...catchallEndpoints)
 
   const filterKeys = Object.keys(filter) as FILTER_FIELD[]
   const filterMapping = definition.filter
-
   for (const filterKey of filterKeys) {
     const filterValue = filter[filterKey]
     if (filterValue && Object.keys(filterMapping[filterKey]).includes(filterValue)) {
-      const registeredEndpoints = filterMapping[filterKey][filterValue]
-      endpoints = endpoints.concat(registeredEndpoints)
+      const filterEndpoints = filterMapping[filterKey][filterValue]
+      endpoints.push(...filterEndpoints)
     }
   }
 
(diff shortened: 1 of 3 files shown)


test/providers/s3-webhook-provider.test.ts
```diff
@@ -1,10 +1,10 @@
-import { S3Client } from '@aws-sdk/client-s3';
-import { WebhookDefinition } from '../../lib/providers/types'
+import { S3Client } from '@aws-sdk/client-s3'
 import { S3WebhookConfigurationProvider } from '../../lib/providers/s3-webhook-provider'
+import { WebhookDefinition } from '../../lib/providers/types'
 
 describe('S3WebhookProvider test', () => {
-  const bucket = 'test-bucket';
-  const key = 'test-key';
```
rileydcampbell, line 7: looks like our linters are fighting? can you undo

lib/providers/json-webhook-provider.ts
```diff
@@ -15,7 +15,7 @@ export class JsonWebhookProvider implements WebhookProvider {
 }
 
 export function findEndpointsMatchingFilter(filter: OrderFilter, definition: WebhookDefinition): Webhook[] {
-  let endpoints: Webhook[] = []
+  let endpoints: Webhook[] = definition['*'] ?? []
```
rileydcampbell, line 18: Just making sure I understand this, is the intention to have a mapping in the definition like ?
```
{
    'filter': {'orderStatus': { 'open': ['mm webhook', ...]}}
    '*' : ['webhook URL that wants every update', ...]
}
```

Mind just adding a short comment for this line?

rileydcampbell: Also, maybe makes more sense to nest the `*` under `filter` field of the `WebhookDefinition`?

rileydcampbell (commented):
Overall, lgtm. Just couple comments.

lib/providers/json-webhook-provider.ts
rileydcampbell: hmmm actually not sure about this^ I think i'm good either way. Just would be nice to add a comment for clarity.

zzmp: I don't want to nest it under `filter` because it is explicitly _not_ filtering. I can add a comment, though.

test/providers/s3-webhook-provider.test.ts
zzmp: I don't think it's that they're fighting, I think it's that this needed linting, and my editor is set up to auto-lint files that I touch. I'd rather leave this, as it looks like it matches the lint of the rest of the repo.

rileydcampbell: ah true, wasn't sure if this repo was `;` or no `;`

rileydcampbell (approved):
lgtm

PR #346, opened 2024-01-22
fix: do not update orders with no status change

zzmp:
Avoids updating orders that have had no status updates.

These updates - which were occurring with every check on the order status - only served to re-trigger downstream dynamodb stream events, which was causing multiple webhooks to be sent for the same open orders.

### DIFF
--- a/lib/handlers/check-order-status/handler.ts
+++ b/lib/handlers/check-order-status/handler.ts
@@ -59,6 +59,7 @@ export class CheckOrderStatusHandler extends SfnLambdaHandler<ContainerInjected,
       provider,
       orderWatcher,
       orderQuoter,
+      orderStatus,
     } = input.requestInjected
 
     const order = checkDefined(
@@ -96,7 +97,9 @@ export class CheckOrderStatusHandler extends SfnLambdaHandler<ContainerInjected,
             gasCostInETH,
             receipt.effectiveGasPrice.toString(),
             receipt.gasUsed.toString(),
-            settledAmounts.reduce((prev, cur) => (prev && BigNumber.from(prev.amountOut).gt(cur.amountOut) ? prev : cur))
+            settledAmounts.reduce((prev, cur) =>
+              prev && BigNumber.from(prev.amountOut).gt(cur.amountOut) ? prev : cur
+            )
           )
 
           const percentDecayed = (timestamp - order.decayStartTime) / (order.decayEndTime - order.decayStartTime)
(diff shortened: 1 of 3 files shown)


lib/handlers/check-order-status/handler.ts
```diff
@@ -306,14 +320,19 @@ export class CheckOrderStatusHandler extends SfnLambdaHandler<ContainerInjected,
...
         getFillLogAttempts,
       },
       'updating order status'
     )
-    await dbInterface.updateOrderStatus(orderHash, orderStatus, txHash, settledAmounts)
+    // Avoid updating the order if the status is unchanged.
+    // This also avoids unnecessarily triggering downstream events from dynamodb changes.
+    if (orderStatus !== lastStatus) {
+      await dbInterface.updateOrderStatus(orderHash, orderStatus, txHash, settledAmounts)
+    }
```
rileydcampbell, line 335: Should we do some kind of short circuit early return instead? Just because otherwise we log all these metrics and debug logs for updating status but then silently just don't actually update it
```if (orderStatus === lastStatus) return```

Comment (zzmp) [reply lib/handlers/check-order-status/handler.ts:335]: I thought we still wanted to return other things, like retryCount. I was trying to not change too much logic, so I didn't read that far into it.

Comment (rileydcampbell) [reply lib/handlers/check-order-status/handler.ts:335]: ah yeah okay so maybe still return that but just avoid logging all the stuff that says we updated the order status
```
if (orderStatus === lastStatus) 
    // all the standard object info we need from this function
    return {
      orderHash: orderHash,
      orderStatus: orderStatus,
      quoteId: quoteId,
      retryCount: retryCount + 1,
      retryWaitSeconds: this.calculateRetryWaitSeconds(chainId, retryCount),
      startingBlockNumber: startingBlockNumber,
      chainId: chainId,
      ...(settledAmounts && { settledAmounts }),
      ...(txHash && { txHash }),
      ...(getFillLogAttempts && { getFillLogAttempts }),
    }
```

File: lib/handlers/check-order-status/handler.ts [hunk 2]
```diff
@@ -306,34 +320,39 @@ export class CheckOrderStatusHandler extends SfnLambdaHandler<ContainerInjected,
         retryCount,
         startingBlockNumber,
         chainId,
+        lastStatus,
```
rileydcampbell, line 323: sorry one more nit, can we also move this log into the conditional?

rileydcampbell (approved):

PR #350, opened 2024-02-01
fix: lowercase getNonce address

marktoda:
Orders are sanitized to have lowercased addresses before pumping into
the DB. This means all `offerers` in the nonce db also have lowercased
addresses. However, the getNonce query does not do sanitization, so if
any integrator requests eip-55 cased nonces, it will not find results in
the db. This commit sanitizes

However, we should hold out for the frontend to make a fix incrementing
the nonce. It's currently relying on nonce randomness to get new nonces


marktoda:
cc @tinaszheng  @just-toby

PR #351, opened 2024-02-02
Add limit get

robert-seitz-uniswap:
adds GET /prod/limit/orders

curl --location '<baseUrl>/prod/limit/orders?orderStatus=open'

response:

```json
{
    "orders": [
        {
            "outputs": [
                {
                    "recipient": "0x5d5577a33abbbae5e7235d2d080caff47e94b7d0",
                    "startAmount": "37563544125610598",
                    "endAmount": "23722713488525370",
                    "token": "0x0000000000000000000000000000000000000000"
                }
            ],
            "encodedOrder": "0x000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000001200000000000000000000000000000000000000000000000000000000065bbc3650000000000000000000000000000000000000000000000000000000065bbc41900000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000a0b86991c6218b36c1d19d4a2e9eb0ce3606eb480000000000000000000000000000000000000000000000000000000005f5e1000000000000000000000000000000000000000000000000000000000005f5e10000000000000000000000000000000000000000000000000000000000000002000000000000000000000000006000da47483062a0d734ba3dc7576ce6a0b645c40000000000000000000000005d5577a33abbbae5e7235d2d080caff47e94b7d032a973db4c1e514de62d1b6be157b4c57dbec049a79c1d921439db70785b45000000000000000000000000000000000000000000000000000000000065bbc425000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000c0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000008573d8281c9a66000000000000000000000000000000000000000000000000005447aea697883a0000000000000000000000005d5577a33abbbae5e7235d2d080caff47e94b7d0",
            "signature": "0x9d6bccaf720382848ae3b6c104fc3312660515bba6a0d15564c9fef31fe479c22443a78aaab730e11497bd8d4183a9aeb1f27af9385911ead334371c04d3f20f1c",
            "input": {
                "endAmount": "100000000",
                "token": "0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48",
                "startAmount": "100000000"
            },
            "orderStatus": "open",
            "createdAt": 1706804100,
            "chainId": 1,
            "orderHash": "0xf64a3de305dc00301a5699f61512bddbc12e1ecee3b79e11785489b3e6429be4",
            "type": "Dutch"
        }
    ]
}
```

### DIFF
--- a/bin/stacks/lambda-stack.ts
+++ b/bin/stacks/lambda-stack.ts
@@ -206,7 +206,7 @@ export class LambdaStack extends cdk.NestedStack {
       runtime: aws_lambda.Runtime.NODEJS_18_X,
       entry: path.join(__dirname, '../../lib/handlers/index.ts'),
       handler: 'getLimitOrdersHandler',
-      timeout: Duration.seconds(29),
+      timeout: Duration.seconds(5),
       memorySize: 512,
       bundling: {
         minify: true,
@@ -378,8 +378,8 @@ export class LambdaStack extends cdk.NestedStack {
 
       const getLimitOrdersTarget = new asg.ScalableTarget(this, `GetLimitOrders-ProvConcASG`, {
         serviceNamespace: asg.ServiceNamespace.LAMBDA,
-        maxCapacity: 50,
-        minCapacity: 1,
+        maxCapacity: provisionedConcurrency * 100,
+        minCapacity: provisionedConcurrency,
         resourceId: `function:${this.getLimitOrdersLambdaAlias.lambda.functionName}:${this.getLimitOrdersLambdaAlias.aliasName}`,
         scalableDimension: 'lambda:function:ProvisionedConcurrency',
       })
--- a/bin/stacks/lambda-stack.ts
+++ b/bin/stacks/lambda-stack.ts
@@ -29,6 +29,7 @@ export class LambdaStack extends cdk.NestedStack {
   public readonly postOrderLambda: aws_lambda_nodejs.NodejsFunction
   public readonly postLimitOrderLambda: aws_lambda_nodejs.NodejsFunction
   public readonly getOrdersLambda: aws_lambda_nodejs.NodejsFunction
+  public readonly getLimitOrdersLambda: aws_lambda_nodejs.NodejsFunction
   public readonly getNonceLambda: aws_lambda_nodejs.NodejsFunction
   private readonly orderNotificationLambda: aws_lambda_nodejs.NodejsFunction
   private readonly getDocsLambda: aws_lambda_nodejs.NodejsFunction
@@ -36,13 +37,11 @@ export class LambdaStack extends cdk.NestedStack {
   public readonly postOrderLambdaAlias: aws_lambda.Alias
   public readonly postLimitOrderLambdaAlias: aws_lambda.Alias
   public readonly getOrdersLambdaAlias: aws_lambda.Alias
+  public readonly getLimitOrdersLambdaAlias: aws_lambda.Alias
   public readonly getNonceLambdaAlias: aws_lambda.Alias
   public readonly getDocsLambdaAlias: aws_lambda.Alias
   public readonly getDocsUILambdaAlias: aws_lambda.Alias
   private readonly orderNotificationLambdaAlias: aws_lambda.Alias
-
-  public readonly getLimitOrdersLambda: aws_lambda_nodejs.NodejsFunction
-  public readonly getLimitOrdersLambdaAlias: aws_lambda.Alias
 
   public readonly chainIdToStatusTrackingStateMachineArn: { [key: string]: string }
   public readonly checkStatusFunction: aws_lambda_nodejs.NodejsFunction
@@ -104,6 +103,12 @@ export class LambdaStack extends cdk.NestedStack {
     this.chainIdToStatusTrackingStateMachineArn = sfnStack.chainIdToStatusTrackingStateMachineArn
     this.checkStatusFunction = sfnStack.checkStatusFunction
 
+    const getOrdersEnv = {
+      stage: props.stage as STAGE,
+      VERSION: '2',
+      NODE_OPTIONS: '--enable-source-maps',
+    }
+
     this.getOrdersLambda = new aws_lambda_nodejs.NodejsFunction(this, `GetOrders${lambdaName}`, {
       role: lambdaRole,
       runtime: aws_lambda.Runtime.NODEJS_18_X,
@@ -115,11 +120,7 @@ export class LambdaStack extends cdk.NestedStack {
         minify: true,
         sourceMap: true,
       },
-      environment: {
-        stage: props.stage as STAGE,
-        VERSION: '2',
-        NODE_OPTIONS: '--enable-source-maps',
-      },
+      environment: getOrdersEnv,
       tracing: aws_lambda.Tracing.ACTIVE,
     })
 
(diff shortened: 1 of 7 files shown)


lib/handlers/get-limit-orders/injector.ts
```diff
@@ -0,0 +1,90 @@
...
+        ...(orderStatus && { orderStatus: orderStatus }),
+        ...(orderHash && { orderHash: orderHash }),
+        ...(offerer && { offerer: offerer.toLowerCase() }),
+        ...(sortKey && { sortKey: sortKey }),
+        ...(filler && { filler: filler.toLowerCase() }),
+        ...(sort && { sort: sort }),
+        ...(chainId && { chainId: chainId }),
+        ...(desc !== undefined && { desc: desc }),
+        ...(orderHashes && { orderHashes: [...new Set(orderHashes)] }),
+      },
```
rileydcampbell, line 84: Can some of this possibly be shared between the limit and dutch endpoint? maybe better to keep separate but just an idea to DRY this out

rileydcampbell (commented):
lgtm, maybe in a follow up PR but let's also update the [swagger file](https://github.com/Uniswap/uniswapx-service/blob/main/swagger.json) at some point for integrators.

lib/handlers/get-limit-orders/injector.ts
robert-seitz-uniswap: I think this should be handled in a future pr. I'm starting a discussion on reorganizing some of this so that future order types can more easily be supported as well, so we will have to identify what can be broken out

bin/stacks/lambda-stack.ts
```diff
@@ -349,6 +376,21 @@ export class LambdaStack extends cdk.NestedStack {
         predefinedMetric: asg.PredefinedMetric.LAMBDA_PROVISIONED_CONCURRENCY_UTILIZATION,
       })
 
+      const getLimitOrdersTarget = new asg.ScalableTarget(this, `GetLimitOrderOrders-ProvConcASG`, {
+        serviceNamespace: asg.ServiceNamespace.LAMBDA,
+        maxCapacity: provisionedConcurrency * 2,
```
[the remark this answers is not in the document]
cgkol: yeap that, or actually @robert-seitz-uniswap feel free to let it go from 1 to 100 or something and it will calibrate itself accordingly

cgkol (commented):
Generally looks good! Generally aligned with Andy's feedback too!

bin/stacks/lambda-stack.ts
```diff
@@ -41,6 +41,9 @@ export class LambdaStack extends cdk.NestedStack {
   public readonly getDocsUILambdaAlias: aws_lambda.Alias
   private readonly orderNotificationLambdaAlias: aws_lambda.Alias
 
+  public readonly getLimitOrdersLambda: aws_lambda_nodejs.NodejsFunction
```
zhongeric, line 44: nit: can we put this up above with the other declarations?

bin/stacks/lambda-stack.ts
```diff
@@ -198,6 +201,24 @@ export class LambdaStack extends cdk.NestedStack {
       tracing: aws_lambda.Tracing.ACTIVE,
     })
 
+    this.getLimitOrdersLambda = new aws_lambda_nodejs.NodejsFunction(this, `GetLimitOrders${lambdaName}`, {
+      role: lambdaRole,
+      runtime: aws_lambda.Runtime.NODEJS_18_X,
+      entry: path.join(__dirname, '../../lib/handlers/index.ts'),
+      handler: 'getLimitOrdersHandler',
+      timeout: Duration.seconds(5),
```
zhongeric, line 209: nit: why timeout 5 seconds here? Seems like 29 up above

bin/stacks/lambda-stack.ts
```diff
@@ -198,6 +201,24 @@ export class LambdaStack extends cdk.NestedStack {
...
+      memorySize: 512,
+      bundling: {
+        minify: true,
+        sourceMap: true,
+      },
+      environment: {
+        stage: props.stage as STAGE,
+        VERSION: '2',
+        NODE_OPTIONS: '--enable-source-maps',
+      },
```
zhongeric, line 219: can we extract this into a constant like above as well?

zhongeric: that way we can pass in prop.envVars too

bin/stacks/lambda-stack.ts
```diff
@@ -349,6 +376,21 @@ export class LambdaStack extends cdk.NestedStack {
         predefinedMetric: asg.PredefinedMetric.LAMBDA_PROVISIONED_CONCURRENCY_UTILIZATION,
       })
 
+      const getLimitOrdersTarget = new asg.ScalableTarget(this, `GetLimitOrders-ProvConcASG`, {
+        serviceNamespace: asg.ServiceNamespace.LAMBDA,
+        maxCapacity: provisionedConcurrency * 100,
```
zhongeric, line 381: nit: 100x seems high? Given getOrders above is 2x . Not sure if I'm missing context here but wanted to flag

lib/handlers/shared/get.ts
```diff
@@ -0,0 +1,90 @@
...
+import { GetOrdersQueryParams, RawGetOrdersQueryParams } from '../get-orders/schema'
+
+export interface ContainerInjected {
+  dbInterface: BaseOrdersRepository
+}
+
+export type GetRequestInjected = {
+  limit: number
+  queryFilters: GetOrdersQueryParams
+  cursor?: string
```
zhongeric, line 16: nit: move optional to end

lib/handlers/shared/get.ts
```diff
@@ -0,0 +1,90 @@
...
+  setGlobalMetrics(metrics)
+
+  return {
+    ...parseGetQueryParams(requestQueryParams),
+    requestId,
+    log,
+  }
+}
+
+export const parseGetQueryParams = (
```
zhongeric, line 57: Is there any request schema validation done before this is called to make sure we're not constructing conflicting queryFilters? I.e. one I can think off is orderHash AND orderHashes

zhongeric (commented):
took a quick look! looking good!

ConjunctiveNormalForm (approved):

bin/stacks/lambda-stack.ts
robert-seitz-uniswap: we're relying on autoscaling to get the number right here. We can think about tightening that up once we get some numbers https://www.notion.so/uniswaplabs/AWS-Lambda-Best-Practices-781917c67edf4dcb96d22e38a6e46411

bin/stacks/lambda-stack.ts
robert-seitz-uniswap: @andysmith415 what would you consider a reasonable timeout. The existing lambdas are at 29 seconds so I'm hesitant to change what's in prod before diving in to the metrics there.

bin/stacks/lambda-stack.ts
robert-seitz-uniswap: which part are you thinking should be passed in envvars exactly? --enable-source-maps?

lib/handlers/shared/get.ts
robert-seitz-uniswap: I don't see any validation in the existing endpoints. It looks like it will return based on the first correct combination it encounters. I'll add a separate validation class in a separate pr.

bin/stacks/lambda-stack.ts
zhongeric: Just the prop.envVars like on line 163

PR #354, opened 2024-02-13
update swagger with limit orders get request

robert-seitz-uniswap:
verified swagger endpoint shows up and generates a valid curl request

### DIFF
--- a/swagger.json
+++ b/swagger.json
@@ -66,6 +66,60 @@
         }
       }
     },
+    "/limit-orders": {
+      "get": {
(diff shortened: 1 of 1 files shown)


cgkol (approved):
Can they use multiple params or only one at a time? I assume multiple but just confirming;

zhongeric (approved):

robert-seitz-uniswap:
> Can they use multiple params or only one at a time? I assume multiple but just confirming;

they can use multiple but it must match one of the indexes

PR #355, opened 2024-02-16
add limit order ddb notifications

robert-seitz-uniswap:


lib/repositories/dutch-orders-repository.ts
```diff
@@ -24,7 +24,7 @@ export class DutchOrdersRepository extends GenericOrdersRepository<string, strin
     })
 
     const orderEntity = new Entity({
-      name: 'Order',
+      name: ORDER_TYPE.DUTCH,
```
cgkol, line 27: why this change btw?

test/unit/providers/s3-webhook-provider.test.ts
```diff
@@ -120,4 +122,57 @@ describe('S3WebhookProvider test', () => {
...
+        let endpoints = findEndpointsMatchingFilter(
+          {
+            filler: '0x1',
+            orderStatus: 'open',
+            offerer: '0x2',
+            orderType: ORDER_TYPE.LIMIT,
+          },
+          mockEndpoints
+        )
+        expect(endpoints).toEqual([{ url: 'webhook.com/0' }])
```
cgkol, line 174: where in the test is the logic defined that `webhook.com/0` is accepting both limit orders and dutch?

robert-seitz-uniswap: it's defined under the * webhook which is what is used as a catchall for the graphql subscription https://github.com/Uniswap/uniswapx-service/blob/5db93307ba2f422f2f38e250d04c58f726122ff2/test/unit/providers/s3-webhook-provider.test.ts#L29

lib/repositories/dutch-orders-repository.ts
robert-seitz-uniswap: I extracted the limit order entity name, so this keeps it consistent. The value is the same

bin/stacks/lambda-stack.ts
```diff
@@ -168,6 +168,17 @@ export class LambdaStack extends cdk.NestedStack {
       })
     )
 
+    this.orderNotificationLambda.addEventSource(
+      new DynamoEventSource(databaseStack.limitOrdersTable, {
+        startingPosition: aws_lambda.StartingPosition.TRIM_HORIZON,
+        batchSize: 1,
+        retryAttempts: 1,
+        bisectBatchOnError: true,
+        reportBatchItemFailures: true,
```
zzmp, line 177: This feels cargo-culted (on both) - it's probably a good spot to optimize should we need to.
You shouldn't change it now, but just to note for the future.

bin/stacks/lambda-stack.ts
```diff
@@ -168,6 +168,17 @@ export class LambdaStack extends cdk.NestedStack {
...
+        onFailure: new SqsDlq(orderNotificationDlq),
```
zzmp, line 178: These options are identical to the ordersTable event source. Can you define it once (`const ordersEventSourceOptions: ... = {`) and re-use it to keep them in sync?

lib/repositories/base.ts
```diff
@@ -1,6 +1,11 @@
 import { OrderEntity, ORDER_STATUS, SettledAmount, SORT_FIELDS } from '../entities/index'
 import { GetOrdersQueryParams } from '../handlers/get-orders/schema'
 
+export const ORDER_TYPE = {
```
zzmp, line 4: It'd be great to export this as a follow-up in uniswapx-sdk (and the whole typing of the webhook).

lib/util/order.ts
```diff
@@ -3,6 +3,7 @@ import { DynamoDBRecord } from 'aws-lambda'
 import { OrderEntity, ORDER_STATUS } from '../entities'
 
 export const DUTCH_LIMIT = 'DutchLimit'
+const DYNAMO_ENTITY_TYPE_FIELD = '_et'
```
zzmp, line 6: Why is this an ad-hoc definition? It's hard to parse, and seems out of place. Is it pulled from some other table definition? Could it be imported to guarantee synchronicity?

lib/providers/json-webhook-provider.ts
```diff
@@ -20,13 +21,22 @@ export function findEndpointsMatchingFilter(filter: OrderFilter, definition: Web
...
-  const filterKeys = Object.keys(filter) as FILTER_FIELD[]
-  const filterMapping = definition.filter
-  for (const filterKey of filterKeys) {
-    const filterValue = filter[filterKey]
-    if (filterValue && Object.keys(filterMapping[filterKey]).includes(filterValue)) {
-      const filterEndpoints = filterMapping[filterKey][filterValue]
-      endpoints.push(...filterEndpoints)
+  /**
+   * ignore limit orders when notifying quoters
+   */
```
zzmp, line 26: This is a jsdoc-comment (with the `/**`), and should only be used to document functions/classes/variables - things that may be exported. For an implementation detail comment, use `/*` or just an inline `//`. This should be linted - it may be worth configuring the linter for this for the future.

this is also strange because it speaks directly to quoters, but quoters are abstracted out of this function.

```suggestion
  // do not notify quoters of limit orders
```

lib/providers/json-webhook-provider.ts
```diff
@@ -20,13 +21,19 @@ export function findEndpointsMatchingFilter(filter: OrderFilter, definition: Web
...
   endpoints.push(...catchallEndpoints)
 
-  const filterKeys = Object.keys(filter) as FILTER_FIELD[]
-  const filterMapping = definition.filter
-  for (const filterKey of filterKeys) {
-    const filterValue = filter[filterKey]
-    if (filterValue && Object.keys(filterMapping[filterKey]).includes(filterValue)) {
-      const filterEndpoints = filterMapping[filterKey][filterValue]
-      endpoints.push(...filterEndpoints)
+  if (filter.orderType !== ORDER_TYPE.LIMIT) {
```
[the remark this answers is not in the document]
zzmp: can you include rationale in the comment? otherwise it's tempting to remove the conditional check, thinking that there's no reason quoters should _not_ be seeing all orders.

zzmp (changes requested):

zzmp (approved):
all of my requests are nits or easily updated, so I'm approving, but please respond/update per the requests :)

lib/util/order.ts
zzmp: > generated from the orm

can you link to documentation in a comment? otherwise it’s impossible to know where that field is coming from, and why it works.

cgkol (approved):

robert-seitz-uniswap:
changed to use uniswapx sdk OrderType

PR #357, opened 2024-02-21
Stop Limit orders from triggering step function

robert-seitz-uniswap:
Step functions were still being created for limit orders. They fail immediately due to being configured to check the dutch order table.

This disables the step function invocation for order types other than dutch orders.

Manual Tests:
Dutch order posts created a step function
Limit order posts did not create a step function


cgkol (approved):
✅

zhongeric (approved):

PR #362, opened 2024-02-26
Use step functions for order status

robert-seitz-uniswap:
use step functions to calculate limit order status
add retry to step function if it exceeds a certain retry threshold
- tracks retryCount on the step function. if it goes over a certain threshold, it will kick off a new step function and cancel end the current one with state orderRetried
- this is to avoid step function limitations on the execution history

testing tonight for launch tomorrow


test/unit/handlers/check-order-status.test.ts
```diff
@@ -219,42 +219,6 @@ describe('Testing check order status handler', () => {
         startingBlockNumber: mockedBlockNumber - FILL_EVENT_LOOKBACK_BLOCKS_ON(1),
       })
     })
-
-    it('should do exponential backoff when retry count > 300', async () => {
```
robert-seitz-uniswap, line 223: these were duped by the service tests

bin/definitions/order-tracking-sfn.json
```diff
@@ -46,9 +51,18 @@
             {
               "Variable": "$.orderStatus",
               "StringEquals": "error"
+            },
+            {
+              "Variable": "$.orderStatus",
+              "StringEquals": "insufficient-funds"
```
cgkol, line 57: fyi: @marktoda we talked about it on slack too but raising visibility

bin/definitions/order-tracking-sfn.json
```diff
@@ -46,9 +51,18 @@
...
             }
           ],
           "Next": "orderInTerminalState"
+        },
+        {
+          "Variable": "$.retryCount",
+          "NumericGreaterThan": 101,
```
cgkol, line 64: lets add a comment so people know what this is for, and that we'll increase it in the future

bin/stacks/lambda-stack.ts
```diff
@@ -94,13 +93,6 @@ export class LambdaStack extends cdk.NestedStack {
       indexCapacityConfig,
     })
 
-    new StatusStack(this, `${SERVICE_NAME}-StatusStack`, {
```
cgkol, line 97: I would comment out here instead and explain what comes up in the future

lib/handlers/shared/sfn.ts
```diff
@@ -0,0 +1,31 @@
...
+  orderHash: string
+  chainId: number
+  orderStatus: ORDER_STATUS
+  quoteId: string
+  orderType: OrderType
+  stateMachineArn: string
+}
+
+//TODO: remove random for sfn name
+const BIG_NUMBER = 100000000000
```
cgkol, line 17: lets make this even bigger. And explain why we have it? I think we can afford at least 10 more zeros 😂

cgkol (commented):
LEts alsso add a comment on the PR that tonight we are following up with more robust tests on this

lib/handlers/post-order/handler.ts
```diff
@@ -145,31 +137,24 @@ export class PostOrderHandler extends APIGLambdaHandler<
...
+        chainId: chainId,
+        orderStatus: ORDER_STATUS.OPEN,
+        quoteId: quoteId ?? '',
+        orderType,
         stateMachineArn,
-        log
-      )
-    }
+      },
+      stateMachineArn
```
cgkol, line 149: passing here `stateMachineArn` twice? is this intentional?

cgkol: i dont think you need the second param if you can pull it from the json right?

cgkol (approved):
Just one small point

PR #369, opened 2024-03-03
Test cleanup: Split tests into unit, integ, e2e + additional mocking

andysmith415:
- Split tests into three buckets: unit (no external dependencies), integration (DDB), end-to-end (going through API Gateway)
- Add a --watch script for unit tests
- Nix the `network-requests` file; it's unused as it stands and we can always bring it back. 
- Create a global mock of aws-embedded-metrics. This can be overwritten on a per-test basis as needed.

Main motivation here is to get the unit tests running more quickly and to get to the bottom of some of the suppressions across the testing codebase, notably the es-lint/ts disable checks as well as the need to have `--detectOpenHandles` and `--forceExits` enabled. The latter is indicative of an unresolved promise firing after the test completes, so something to figure out.

The check-order-status tests have some dangling promises that are prevently clean exits. Until those are addressed, they can live in integ/ where `--forceExits` is used.

### DIFF
--- a/bin/stacks/lambda-stack.ts
+++ b/bin/stacks/lambda-stack.ts
@@ -298,6 +298,12 @@ export class LambdaStack extends cdk.NestedStack {
         filterPattern: '{ $.eventType = "OrderPosted" }',
         logGroupName: this.postOrderLambda.logGroup.logGroupName,
       })
+
+      new cdk.aws_logs.CfnSubscriptionFilter(this, 'PostedLimitOrderSub', {
+        destinationArn: props.envVars['POSTED_ORDER_DESTINATION_ARN'],
+        filterPattern: '{ $.eventType = "OrderPosted" }',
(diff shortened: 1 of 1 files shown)


robert-seitz-uniswap (approved):

__mocks__/aws-embedded-metrics.ts
```diff
@@ -0,0 +1,15 @@
+const original = jest.requireActual('aws-embedded-metrics')
```
cgkol, line 1: n00b question but how is this passed in on the test, aka how do the tests know to read all exports from __mocks__ directory?

cgkol (approved):

__mocks__/aws-embedded-metrics.ts
cgkol: thank you sir

marktoda (approved):

PR #371, opened 2024-03-03
Silo handler functions

andysmith415:
Move handlers from top-level index.ts to their own index.ts file. Two reasons to do this:
1. It allows the lambdas to only import the code that they use, reducing the bundle size and speeding up the deployments by a bit. 
2. It allows lambda-specific configuration to be colocated with handler instantiation (coming soon).


robert-seitz-uniswap (approved):

cgkol (approved):
nice ✅

PR #372, opened 2024-03-04
refactor fillevent check

robert-seitz-uniswap:
fillEvent check is taking 10-20s so push it down to where it's needed


cgkol (commented):
So essnetially only use it for these two cases is that right? expired and nonceUsed?

robert-seitz-uniswap:
yes instead of checking it for all cases

PR #377, opened 2024-03-07
Extract post-order-handler body into service

andysmith415:
Pull the order handler logic into a separate "Service" class. This class preserves most of the existing behavior with some slight renames. It throws named exceptions that are mapped to HTTP status codes in the handler.

Future iterations will properly inject this dependency. Some base classes will fall out of this that can be used across the order types.

### DIFF
--- a/lib/HttpStatusCode.ts
+++ b/lib/HttpStatusCode.ts
@@ -1,0 +1,13 @@
+export enum HttpStatusCode {
+  Ok = 200,
+  Created = 201,
+  BadRequest = 400,
+  Unauthorized = 401,
+  Forbidden = 403,
+  NotFound = 404,
+  Conflict = 409,
+  InternalServerError = 500,
+  NotImplemented = 501,
+  BadGateway = 502,
+  ServiceUnavailable = 503,
+}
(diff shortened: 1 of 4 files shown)


lib/services/UniswapXOrderService.ts
```diff
@@ -0,0 +1,128 @@
...
+    })
+  }
+
+  private async startOrderTracker(
+    orderHash: string,
+    chainId: number,
+    quoteId: string | undefined,
+    orderType: OrderType
+  ) {
+    const stateMachineArn = checkDefined(process.env[`STATE_MACHINE_ARN_${chainId}`])
```
robert-seitz-uniswap, line 115: move this up to DI level eventually

robert-seitz-uniswap (approved):

lib/services/UniswapXOrderService.ts
```diff
@@ -0,0 +1,128 @@
...
+  constructor(
+    private readonly orderValidator: OffChainOrderValidator,
+    private readonly onChainValidator: OnChainOrderValidator,
+    private readonly repository: BaseOrdersRepository,
+    private readonly logger: Logger,
+    private readonly getMaxOpenOrders: (offerer: string) => number
+  ) {}
+
+  async createOrder(
+    order: DutchOrder,
```
robert-seitz-uniswap, line 25: are we going to have this service handle a generic order or have a service per type

robert-seitz-uniswap (approved):

PR #380, opened 2024-03-07
Extract order validator

andysmith415:
- Add a top-level Config file with checked environment variables. The idea here is that if we require environment variables to be set, we should fail as soon as possible vs needing to have that specific codepath run. 
- Inject the OnChainOrdervalidators into the constructor of the Post order + Post limit order handlers

Beta + Prod have secrets correctly set for Mainnet, Polygon, Goerli.

### DIFF
--- a/lib/handlers/post-order/injector.ts
+++ b/lib/handlers/post-order/injector.ts
@@ -1,12 +1,10 @@
-import { OrderType, OrderValidator as OnchainValidator } from '@uniswap/uniswapx-sdk'
+import { OrderType } from '@uniswap/uniswapx-sdk'
 import { MetricsLogger } from 'aws-embedded-metrics'
 import { APIGatewayEvent, Context } from 'aws-lambda'
 import { DynamoDB } from 'aws-sdk'
 import { default as Logger } from 'bunyan'
-import { ethers } from 'ethers'
 import { BaseOrdersRepository } from '../../repositories/base'
 import { DutchOrdersRepository } from '../../repositories/dutch-orders-repository'
-import { SUPPORTED_CHAINS } from '../../util/chain'
 import { ONE_DAY_IN_SECONDS } from '../../util/constants'
 import { setGlobalLogger } from '../../util/log'
 import { setGlobalMetrics } from '../../util/metrics'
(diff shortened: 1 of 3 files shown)


lib/handlers/post-limit-order/index.ts
```diff
@@ -1,8 +1,26 @@
+import { OrderValidator } from '@uniswap/uniswapx-sdk'
+import { ethers } from 'ethers'
+import { CONFIG } from '../../Config'
+import { SUPPORTED_CHAINS } from '../../util/chain'
+import { OnChainValidatorMap } from '../OnChainValidatorMap'
 import { PostOrderHandler } from '../post-order/handler'
 import { PostLimitOrderInjector } from './injector'
 
+const onChainValidatorMap = new OnChainValidatorMap()
```
robert-seitz-uniswap, line 9: would move this to utils or somewhere global

lib/handlers/post-limit-order/index.ts
```diff
@@ -1,8 +1,26 @@
...
+
+for (const chainId of SUPPORTED_CHAINS) {
```
robert-seitz-uniswap, line 11: why not pull supported chains into the constructor of onChainValidatorMap

lib/handlers/post-order/handler.ts
```diff
@@ -18,37 +27,30 @@ export class PostOrderHandler extends APIGLambdaHandler<
...
+      containerInjected: { dbInterface, orderValidator, orderType, getMaxOpenOrders },
     } = params
 
     log.info('Handling POST order request', params)
     log.info(
       {
-        onchainValidatorByChainId: Object.keys(onchainValidatorByChainId).map(
-          (chainId) => onchainValidatorByChainId[Number(chainId)].orderQuoterAddress
-        ),
+        onchainValidatorByChainId: this.onChainValidatorMap.debug(),
```
robert-seitz-uniswap, line 50: you want to leave this in?

lib/handlers/post-order/injector.ts
```diff
@@ -18,30 +16,16 @@ import { PostOrderRequestBody } from './schema'
...
   orderValidator: OrderValidator
-  onchainValidatorByChainId: { [chainId: number]: OnchainValidator }
   orderType: OrderType
   // return the number of open orders the given offerer is allowed to have at a time
   getMaxOpenOrders: (offerer: string) => number
 }
 
 export class PostOrderInjector extends ApiInjector<ContainerInjected, ApiRInj, PostOrderRequestBody, void> {
   public async buildContainerInjected(): Promise<ContainerInjected> {
-    const onchainValidatorByChainId: { [chainId: number]: OnchainValidator } = {}
```
robert-seitz-uniswap, line 29: nice

robert-seitz-uniswap (approved):

PR #384, opened 2024-03-08
Inject UniswapXService

andysmith415:
This PR pulls the remaining dependencies out of containerInjected for the post order endpoints and injects a UniswapXService into the constructor. The idea here is once we have the service injected, we can start to generalize it for multiple order types.

There is some hackery around logging that is unavoidable in its current state. This stems from the fact that adding context to the logger (in our case, the AWS requestID) requires creating a new instance of the logger, which means that you cannot inject the logger that will be used at the time of handler construction; you need to set this across the attributes. 

This can be solved by moving over to powertools logger or doing some container rebinding at request time. Both are viable follow-ups.


robert-seitz-uniswap:
this can use the AnalyticsService and remove the bunyan log now

robert-seitz-uniswap (approved):
Nice

PR #388, opened 2024-03-12
Deadline validity period error

robert-seitz-uniswap:
fix Deadline Validity period error message to reflect actual deadline

### DIFF
--- a/lib/util/order-validator.ts
+++ b/lib/util/order-validator.ts
@@ -15,7 +15,7 @@ export type SkipValidationMap = {
 export class OrderValidator {
   constructor(
     private readonly getCurrentTime: () => number,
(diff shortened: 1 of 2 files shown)


lib/util/order-validator.ts
```diff
@@ -121,7 +121,9 @@ export class OrderValidator {
     if (deadline > this.getCurrentTime() + this.deadlineValidityPeriod) {
       return {
         valid: false,
-        errorString: `Deadline field invalid: Order expiry cannot be larger than thirty minutes`,
+        errorString: `Deadline field invalid: Order expiry cannot be larger than ${
+          this.deadlineValidityPeriod / 60
```
robert-seitz-uniswap, line 125: deadlinValidityPeriod is in seconds

robert-seitz-uniswap: just renamed it

PR #392, opened 2024-03-13
Generalized order parsing

andysmith415:
This PR lays the groundwork for the generalized POST order endpoint. It's a bit beefy but mostly tests. 

A few changes here:

- Makes the request body take in either the existing payload sans OrderType or an explicitly typed payload. 
- Create an internal representation of an Order that wraps the UniswapSDK. The idea is that we then have services that can take in the generic Order model and can act on them. This is a nice way to avoid having a method that takes in the superset of all possible attributes
- Create a "PostOrderParser" that can take in the Post request and parse the Order, throwing where appropriate
- Adds DutchV2Order + RelayOrder factories - tests for DutchV2Order skipped until https://github.com/Uniswap/uniswapx-sdk/pull/124 lands

### DIFF
--- a/lib/crons/gs-reaper.ts
+++ b/lib/crons/gs-reaper.ts
@@ -3,7 +3,7 @@ import { DynamoDB } from 'aws-sdk'
 import { default as bunyan, default as Logger } from 'bunyan'
 
 import { metricScope, MetricsLogger, Unit } from 'aws-embedded-metrics'
-import { ORDER_STATUS } from '../entities'
+import { OrderEntity, ORDER_STATUS } from '../entities'
 import { BaseOrdersRepository } from '../repositories/base'
 import { DutchOrdersRepository } from '../repositories/dutch-orders-repository'
 import { DYNAMO_BATCH_WRITE_MAX, ONE_HOUR_IN_SECONDS } from '../util/constants'
@@ -25,7 +25,7 @@ async function main(metrics: MetricsLogger) {
 }
 
 export async function deleteStaleOrders(
-  repo: BaseOrdersRepository,
+  repo: BaseOrdersRepository<OrderEntity>,
   log: Logger,
   metrics?: MetricsLogger
 ): Promise<void> {
--- a/lib/entities/Order.ts
+++ b/lib/entities/Order.ts
@@ -54,6 +54,35 @@ export type OrderEntity = {
   settledAmounts?: SettledAmount[]
 }
 
+export type RelayOrderEntity = {
+  type: OrderType
+  encodedOrder: string
+  signature: string
+  nonce: string
+  orderHash: string
+  orderStatus: ORDER_STATUS
+  chainId: number
+  offerer: string
+  reactor: string
+  decayStartTime: number
+  decayEndTime: number
+  deadline: number
+  input: {
+    token: string
+    amount: string
+    recipient: string
+  }
+  outputs: OrderOutput[]
+  // Filler field is defined when the order has been filled and the status tracking function has recorded the filler address.
+  filler?: string
+  // QuoteId field is defined when the order has a quote associated with it.
+  quoteId?: string
+  // TxHash field is defined when the order has been filled and there is a txHash associated with the fill.
+  txHash?: string
+  // SettledAmount field is defined when the order has been filled and the fill amounts have been recorded.
+  settledAmounts?: SettledAmount[]
+}
+
 export enum SORT_FIELDS {
   CREATED_AT = 'createdAt',
 }
--- a/lib/handlers/check-order-status/handler.ts
+++ b/lib/handlers/check-order-status/handler.ts
@@ -2,6 +2,7 @@ import { MetricUnits } from '@aws-lambda-powertools/metrics'
 import { OrderType } from '@uniswap/uniswapx-sdk'
 import { DynamoDB } from 'aws-sdk'
 import Joi from 'joi'
+import { OrderEntity } from '../../entities'
 import { CheckOrderStatusHandlerMetricNames, powertoolsMetric } from '../../Metrics'
 import { BaseOrdersRepository } from '../../repositories/base'
 import { LimitOrdersRepository } from '../../repositories/limit-orders-repository'
@@ -18,7 +19,7 @@ export class CheckOrderStatusHandler extends SfnLambdaHandler<ContainerInjected,
   private _checkLimitOrderStatusService!: CheckOrderStatusService
 
   // TODO: Inject this
-  private getCheckOrderStatusService(dbInterface: BaseOrdersRepository) {
+  private getCheckOrderStatusService(dbInterface: BaseOrdersRepository<OrderEntity>) {
     if (!this._checkOrderStatusService) {
       this._checkOrderStatusService = new CheckOrderStatusService(
         dbInterface,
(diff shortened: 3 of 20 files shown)


lib/handlers/post-order/PostOrderBodyParser.ts
```diff
@@ -0,0 +1,145 @@
...
+        return this.tryParseDutchV1Order(encodedOrder, signature, chainId, body.quoteId)
+      case OrderType.Limit:
+        return this.tryParseLimitOrder(body.encodedOrder, body.signature, body.chainId, body.quoteId)
+      case OrderType.Dutch_V2:
+        return this.tryParseDutchV2Order(encodedOrder, signature, chainId)
+
+      case OrderType.Relay:
+        return this.tryParseRelayOrder(encodedOrder, signature, chainId)
+
+      case undefined:
```
robert-seitz-uniswap, line 36: This feels a bit weird. Not sure if it would be better just return undefined here and expose functions to try to parse the legacy path. Realistically I don't know when we wouldn't have the type based on the new endpoint requiring it or the old endpoints implicitly having it in the route

lib/handlers/post-order/PostOrderBodyParser.ts
```diff
@@ -0,0 +1,145 @@
...
+} from '@uniswap/uniswapx-sdk'
+import { UnexpectedOrderTypeError } from '../../errors/UnexpectedOrderTypeError'
+import { DutchV1Order } from '../../models/DutchV1Order'
+import { DutchV2Order } from '../../models/DutchV2Order'
+import { LimitOrder } from '../../models/LimitOrder'
+import { Order } from '../../models/Order'
+import { RelayOrder } from '../../models/RelayOrder'
+import { PostOrderRequestBody } from './schema'
+
+export class PostOrderBodyParser {
```
robert-seitz-uniswap, line 18: Is this not a generic orderParser? Is it specific to the Post body?

lib/handlers/post-order/PostOrderBodyParser.ts
```diff
@@ -0,0 +1,145 @@
...
+        return this.tryParseDutchOrder(encodedOrder, signature, chainId, body.quoteId)
+      default:
+        throw new Error(`No parser configured for orderType ${orderType}`)
+    }
+  }
+
+  private tryParseRelayOrder(encodedOrder: string, signature: string, chainId: number): RelayOrder {
+    try {
+      const order = this.relayParser.parseOrder(encodedOrder, chainId)
+      const orderType = this.relayParser.getOrderType(order)
```
robert-seitz-uniswap, line 48: wouldn't line 47 throw if line 48 is not of type Relay?

lib/models/Order.ts
```diff
@@ -0,0 +1,5 @@
+import { OrderType } from '@uniswap/uniswapx-sdk'
+
+export abstract class Order {
```
robert-seitz-uniswap, line 3: should we add signature, encodedOrder, any other minimally required fields

robert-seitz-uniswap:
this looks good

robert-seitz-uniswap (approved):

lib/handlers/post-order/PostOrderBodyParser.ts
```diff
@@ -0,0 +1,142 @@
...
+      if (orderType === OrderType.Relay) {
+        return new RelayOrder(order as SDKRelayOrder, signature, chainId)
+      }
+      throw new UnexpectedOrderTypeError(orderType)
+    } catch (err) {
```
zhongeric, line 50: ok so the explicit error that will be thrown is something along the lines of `no reactor address mapping found for order type` since if we are parsing a wrong order the parsed reactor address will not be correct. I would prefer us catching for that explicitly instead of all exceptions

lib/handlers/post-order/PostOrderBodyParser.ts
```diff
@@ -0,0 +1,142 @@
...
+        err,
+        encodedOrder,
+        chainId,
+        signature,
+      })
+      throw err
+    }
+  }
+
+  tryParseDutchOrder(encodedOrder: string, signature: string, chainId: number, quoteId?: string) {
```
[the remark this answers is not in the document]
robert-seitz-uniswap: wouldn't some other error come up when decoding in that case since the output vs expected output will be different?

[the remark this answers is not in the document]
zhongeric: So the way we determine the order type of orders is by checking the encoded address at some byte offset. Then with the orderType we know the decoding schema which gets us the outputted order. However, what I'm saying is that an order can be maliciously crafted, especially since some fields can be dynamic, where it may be able to cleverly encode the reactor address of another reactor at some byte offset, tricking us into parsing it as a different order

lib/handlers/post-order/PostOrderBodyParser.ts
```diff
@@ -0,0 +1,142 @@
...
+        return this.tryParseLimitOrder(encodedOrder, signature, chainId, body.quoteId)
+      case OrderType.Dutch_V2:
+        return this.tryParseDutchV2Order(encodedOrder, signature, chainId)
+      case OrderType.Relay:
+        return this.tryParseRelayOrder(encodedOrder, signature, chainId)
+
+      case undefined:
+        // If an OrderType is not explicitly set, it is the legacy format which is either a DutchOrderV1 or a LimitOrder.
+        // Try to parse both and see which hits.
+        return this.tryParseDutchOrder(encodedOrder, signature, chainId, body.quoteId)
```
cgkol, line 38: so this is strictly for backwards compatibility right?

lib/handlers/post-order/PostOrderBodyParser.ts
cgkol: i'll pile on to this, you can never get here otherwise because this is only called when the type is such, but we might change it in the future i guess so we can keep it. Defensive programming!

lib/handlers/post-order/PostOrderBodyParser.ts
```diff
@@ -0,0 +1,142 @@
...
+        err,
+        encodedOrder,
+        chainId,
+        signature,
+      })
+      throw err
+    }
+  }
+
+  private tryParseDutchV2Order(encodedOrder: string, signature: string, chainId: number): DutchV2Order {
```
cgkol, line 84: nit: btw all these functions are boilerplate code, we could probably somehow do it with ? and types but probably not worth the apin

lib/handlers/post-order/PostOrderBodyParser.ts
```diff
@@ -0,0 +1,142 @@
...
+        // If an OrderType is not explicitly set, it is the legacy format which is either a DutchOrderV1 or a LimitOrder.
+        // Try to parse both and see which hits.
+        return this.tryParseDutchOrder(encodedOrder, signature, chainId, body.quoteId)
+    }
+  }
+
+  private tryParseRelayOrder(encodedOrder: string, signature: string, chainId: number): RelayOrder {
+    try {
+      const order = this.relayParser.parseOrder(encodedOrder, chainId)
+      const orderType = this.relayParser.getOrderType(order)
```
cgkol, line 45: why do you need the parser given that you have explicit function for it?

cgkol (commented):
Looks good generally, some questions

zhongeric (approved):

PR #396, opened 2024-03-14
Order Dispatcher

andysmith415:
Branches off of https://github.com/Uniswap/uniswapx-service/pull/392

Create a generic "OrderDispatcher" that can be used to take a generic order / query and fire it off to the corresponding service. This lets the dispatcher be the primary entrypoint for create / listing orders so the handlers don't have to deal with the specifics of the interactions. 

In the future, this will house a RelayOrderService and a CrossChainOrderService.

### DIFF
--- a/test/unit/services/UniswapXOrderService.test.ts
+++ b/test/unit/services/UniswapXOrderService.test.ts
@@ -1,3 +1,0 @@
-describe('UniswapXOrderService', () => {
-  it('does stuff', () => {})
-})


robert-seitz-uniswap (approved):

cgkol (commented):
I think this is fine fwiw, but it does feel like an extra abstraction without any convolutd logic in it. I see in your PR description that we will add more to it down the road?

PR #402, opened 2024-03-20
bump time to check expiration in e2e tests

robert-seitz-uniswap:
due to how we update the expiration, it might not be updated to expired within 1 second, so bumping the wait time by 20 seconds


cgkol (approved):
these take will take even longer now 😅

PR #406, opened 2024-03-28
chore(analytics): add new subscription filter for insufficient-funds logs

ConjunctiveNormalForm:


robert-seitz-uniswap:
just to note here: seems like there is a limit of 5 regex subscriptions per log group and this would bring us to 2 subscriptions on `checkStatusFunction.logGroup.logGroupName`

source: https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/FilterAndPatternSyntax.html

robert-seitz-uniswap (approved):

ConjunctiveNormalForm:
> just to note here: seems like there is a limit of 5 regex subscriptions per log group and this would bring us to 2 subscriptions on `checkStatusFunction.logGroup.logGroupName`
> 
> source: https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/FilterAndPatternSyntax.html

Yep, and that's up from max of 3 when we first implemented this haha

PR #438, opened 2024-05-17
fix: add requestId to dynamo-toolbox order entity

ConjunctiveNormalForm:
Without adding `requestId`, the updated unit tests would fail with `Field 'requestId' does not have a mapping or alias` matching what we see in prod


zhongeric (approved):
thank you!

PR #439, opened 2024-05-30
Add E2E test for pipeline check in beta

codyborn:
Adding an E2E test to ensure the services are working together before rolling out to production.

To make this test reliable I:
- Force open orders (don't need to worry about empty quotes)
- Decrease the amount prior to order so that the order is valid but wouldn't be filled

Before merging we need to add the following environment variables to Beta secrets
```
URA_SERVICE_URL="https://..."
GPA_SERVICE_URL="https://..."
COSIGNER_ADDRESS="..."
```

### DIFF
--- a/test/e2e/order.test.ts
+++ b/test/e2e/order.test.ts
@@ -1,5 +1,5 @@
 /* eslint-disable jest/no-disabled-tests */
-import { CosignedV2DutchOrder, CosignerData, DutchOrder, DutchOrderBuilder, OrderType, REACTOR_ADDRESS_MAPPING, SignedUniswapXOrder, UnsignedV2DutchOrder, V2DutchOrderBuilder } from '@uniswap/uniswapx-sdk'
+import { DutchOrder, DutchOrderBuilder, REACTOR_ADDRESS_MAPPING, SignedUniswapXOrder, UnsignedV2DutchOrder, V2DutchOrderBuilder } from '@uniswap/uniswapx-sdk'
 import { factories } from '@uniswap/uniswapx-sdk/dist/src/contracts/index'
 import axios from 'axios'
 import dotenv from 'dotenv'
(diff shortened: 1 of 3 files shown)


test/e2e/order.test.ts
```diff
@@ -305,7 +433,7 @@ describe.skip('/dutch-auction/order', () => {
     return receipt.transactionHash
   }
 
-  describe('orders endpoint sanity checks', () => {
+  describe.skip('orders endpoint sanity checks', () => {
```
ConjunctiveNormalForm, line 436: why do we skip those tests now?

test/e2e/order.test.ts
```diff
@@ -354,34 +482,39 @@ describe.skip('/dutch-auction/order', () => {
   })
 
   describe('order endpoint sanity checks', () => {
-    it('2xx', async () => {
-      await buildAndSubmitOrder(aliceAddress, amount, DEFAULT_DEADLINE_SECONDS, WETH_GOERLI, UNI_GOERLI)
+    it.skip('2xx', async () => {
+      await buildAndSubmitOrder(aliceAddress, amount, DEFAULT_DEADLINE_SECONDS, wethAddress, uniAddress)
+    })
+
+    it('2xx with an order from URA', async () => {
```
ConjunctiveNormalForm, line 489: nit: maybe move this closer to top of file? took me a while to find the case that isn't skipped haha

ConjunctiveNormalForm (approved):

test/e2e/order.test.ts
codyborn: Previously the entire block was skipped with line 43:
`describe.skip('/dutch-auction/order'`
so now that I'm not skipping it, I need to skip the individual test cases

test/e2e/order.test.ts
codyborn: good point

PR #441, opened 2024-05-31
feat: add nonce to GET /orders response for both v1 and v2 orders

ConjunctiveNormalForm:
completes [PROTO-255](https://linear.app/uniswap/issue/PROTO-255/orderservice-nonce-not-part-of-get-orders-response)

verified working locally
Screenshot 2024-05-31 at 4 19 14 PM


codyborn (approved):
🚢

PR #452, opened 2024-07-23
feat: add PriorityOrder support at DB level and in get-orders lambda

ConjunctiveNormalForm:
Ended up not adding a new index for `orderType`s, but kept the existing implementation which is `query + filter expressions`.

Tested locally
Screenshot 2024-07-24 at 2 20 01 PM

Will add support in post-orders in a separate PR.


lib/entities/Order.ts
```diff
@@ -56,16 +62,60 @@ export type UniswapXOrderEntity = {
...
+// Db representation of Dutch V1, Dutch V2, or Limit Order
+// indexes are returned at runtime but not represented on this type. Ideally we will include a mapping at repo layer boundary
+export type UniswapXOrderEntity = DutchV1OrderEntity | DutchV2OrderEntity | PriorityOrderEntity
+
 export enum SORT_FIELDS {
   CREATED_AT = 'createdAt',
 }
+
+export function isPriorityOrderEntity(order: UniswapXOrderEntity): order is PriorityOrderEntity {
+  return 'cosignerData' in order && 'auctionTargetBlock' in order.cosignerData
```
zhongeric, line 112: Can we use the `type` field for the type coercion?

lib/handlers/get-orders/schema/GetPriorityOrderResponse.ts
```diff
@@ -0,0 +1,73 @@
...
+  settledAmounts: Joi.array().items(
+    Joi.object({
+      tokenOut: FieldValidator.isValidEthAddress(),
+      amountOut: FieldValidator.isValidAmount(),
+      tokenIn: FieldValidator.isValidEthAddress(),
+      amountIn: FieldValidator.isValidAmount(),
+    })
+  ),
+  auctionStartBlock: Joi.number().min(0),
+  baselinePriorityFeeWei: Joi.string().regex(/^[0-9]+$/),
```
zhongeric, line 66: Can we use `isValidAmount` here?

lib/repositories/generic-orders-repository.ts
```diff
@@ -53,6 +53,17 @@ export abstract class GenericOrdersRepository<
     return await this.queryOrderEntity(orderStatus, TABLE_KEY.ORDER_STATUS, limit, cursor, sortKey, sort, desc)
   }
 
+  //public async getByOrderType(
```
zhongeric, line 56: remove?

lib/services/analytics-service.ts
```diff
@@ -86,9 +107,10 @@ export class AnalyticsService implements AnalyticsServiceInterface {
         orderHash: fill.orderHash,
         orderType: order.type,
         quoteId: quoteId,
-        exclusiveFiller: (order as UniswapXOrderEntity).cosignerData ?
-          this.getFillerAddress((order as UniswapXOrderEntity).cosignerData?.exclusiveFiller ?? AddressZero):
-          AddressZero,
+        exclusiveFiller:
+          'cosignerData' in order && 'exclusiveFiller' in order.cosignerData
```
zhongeric, line 111: could we do `isDutchV2OrderEntity()` here?

test/factories/SDKPriorityOrderFactory.test.ts
```diff
@@ -0,0 +1,7 @@
+import { SDKPriorityOrderFactory } from './SDKPriorityOrderFactory'
+
+describe('SDKDutchOrderV2Factory', () => {
```
zhongeric, line 3: nit: naming here

zhongeric (approved):

lib/services/analytics-service.ts
ConjunctiveNormalForm: unfortunately not here because order could be RelayOrderEntity type

PR #453, opened 2024-07-24
fix: build errors

ConjunctiveNormalForm:
fixing some build errors in #452


lib/util/order.ts
```diff
@@ -48,7 +48,7 @@ export const eventRecordToOrder = (record: DynamoDBRecord): ParsedOrder => {
 export const formatOrderEntity = (
   decodedOrder: DutchOrder,
   signature: string,
-  orderType: OrderType.Dutch | OrderType.Dutch_V2 | OrderType.Limit,
```
zhongeric, line 51: do we need to special case DutchV2 here then to pull the start/end times from the cosigner Data?

test/integ/repositories/dynamo-repository.test.ts
```diff
@@ -174,7 +174,7 @@ describe('OrdersRepository put item test', () => {
   it('should succesfully put a priority order in table', async () => {
     expect(() => {
       mockTime(1)
-      ordersRepository.putOrderAndUpdateNonceTransaction(PRIORITY_ORDER_1 as UniswapXOrderEntity)
+      ordersRepository.putOrderAndUpdateNonceTransaction(PRIORITY_ORDER_1 as unknown as UniswapXOrderEntity)
```
zhongeric, line 177: I don't love the unknown cast - why doesn't PriorityOrder fit into UniswapXOrderEntity?

zhongeric (approved):

lib/util/order.ts
ConjunctiveNormalForm: this function is only being used for v1 orders (and limit, which is also v1) so probably no need

test/integ/repositories/dynamo-repository.test.ts
ConjunctiveNormalForm: discussed offline, all mock orders only contain a subset of actual order entity type fields and that's fine for testing purpose

PR #466, opened 2024-08-01
chore: priority cosigner override

ConjunctiveNormalForm:
set `auctionTargetBlock` to current block + 1

### DIFF
--- a/lib/services/UniswapXOrderService.ts
+++ b/lib/services/UniswapXOrderService.ts
@@ -19,7 +19,6 @@ import { GetDutchV2OrderResponse } from '../handlers/get-orders/schema/GetDutchV
 import { GetOrdersResponse } from '../handlers/get-orders/schema/GetOrdersResponse'
 import { GetPriorityOrderResponse } from '../handlers/get-orders/schema/GetPriorityOrderResponse'
 import { OnChainValidatorMap } from '../handlers/OnChainValidatorMap'
-import { Cosigner } from '../handlers/post-order'
 import { kickoffOrderTrackingSfn } from '../handlers/shared/sfn'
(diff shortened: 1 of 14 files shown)


lib/services/UniswapXOrderService.ts
```diff
@@ -49,9 +53,12 @@ export class UniswapXOrderService {
       await this.validateOrder(order.inner, order.signature, order.chainId)
       orderEntity = order.toEntity(ORDER_STATUS.OPEN)
     } else if (order instanceof PriorityOrder) {
-      // TODO: do cosigner overriding here
       await this.validateOrder(order.inner, order.signature, order.chainId)
-      orderEntity = order.toEntity(ORDER_STATUS.OPEN)
+      const kmsKeyId = checkDefined(process.env.KMS_KEY_ID, 'KMS_KEY_ID is not defined')
+      const awsRegion = checkDefined(process.env.REGION, 'REGION is not defined')
+      const cosigner = new KmsSigner(new KMSClient({ region: awsRegion }), kmsKeyId)
```
zhongeric, line 59: Is there a specific reason why we are initializing the kms signer here instead of passing it into the class as a depedency? Feels odd to have this logic within `createOrder` that throws if env vars are not defined

zhongeric (approved):
Looks great, just one comment

lib/services/UniswapXOrderService.ts
ConjunctiveNormalForm: yeah, following what we do in param-api here https://github.com/Uniswap/uniswapx-parameterization-api/pull/358

zhongeric (approved):

PR #468, opened 2024-08-05
chore: add priority order cosigner check

ConjunctiveNormalForm:


zhongeric (approved):
lgtm. though we should make sure to have URA and order service use the same secret arns

zhongeric (approved):

PR #480, opened 2024-08-14
fix: parse priority order settled amounts

ConjunctiveNormalForm:


lib/handlers/check-order-status/service.ts
```diff
@@ -126,8 +126,13 @@ export class CheckOrderStatusService {
...
           fillEvent,
-          block.timestamp,
-          parsedOrder as DutchOrder | CosignedV2DutchOrder
+          {
+            timestamp: block.timestamp,
+            gasPrice: tx.gasPrice,
+            maxPriorityFeePerGas: tx.maxPriorityFeePerGas,
+            maxFeePerGas: tx.maxFeePerGas,
+          },
+          parsedOrder as DutchOrder | CosignedV2DutchOrder | CosignedPriorityOrder
```
zhongeric, line 135: could we define a unioned type and use it here? there may be something in the SDK, or could do like `AllowableOrderTypes` since order service doesn't accept all order types

lib/handlers/check-order-status/util.ts
```diff
@@ -9,16 +10,96 @@ import {
...
+  const nativeOutputs = parsedOrder.info.outputs.filter((output) => output.token.toLowerCase() === NATIVE_ADDRESS)
+  const settledAmounts: SettledAmount[] = []
+  let amountIn: string
+
+  // exact_output
+  if (parsedOrder.info.input.mpsPerPriorityFeeWei.eq(0)) {
+    amountIn = parsedOrder.info.input.amount.toString()
+
+    // TODO: gracefully handle undefined maxPriorityFeePerGas - does this ever happen?
+    const resolvedOrder = parsedOrder.resolve({ priorityFee: metadata.maxPriorityFeePerGas! })
```
zhongeric, line 53: if the txn is not 1559 i think this will be undefined and throw, we can just set it equal to 0 here or throw with a custom error?

lib/handlers/check-order-status/util.ts
```diff
@@ -9,16 +10,96 @@ import {
...
+export function getPrioritySettledAmounts(
+  fill: FillInfo,
+  metadata: FillMetadata,
+  parsedOrder: CosignedPriorityOrder
+): SettledAmount[] {
+  const nativeOutputs = parsedOrder.info.outputs.filter((output) => output.token.toLowerCase() === NATIVE_ADDRESS)
+  const settledAmounts: SettledAmount[] = []
+  let amountIn: string
+
+  // exact_output
```
zhongeric, line 48: *exact input

test/integ/handlers/check-order-status.test.ts
```diff
@@ -422,152 +423,374 @@ describe('Testing check order status handler', () => {
...
-        [
+        mockOrder.resolve = () => {
+          return {
+            input: { token: resolvedInput.token, amount: resolvedInput.amount },
+            outputs: [resolvedOutput],
+          }
+        }
+        const mockFillInfo = getMockFillInfo([resolvedInput], [])
+
+        const settledAmounts = getSettledAmounts(mockFillInfo, { timestamp: 100 }, mockOrder)
```
zhongeric, line 558: can we add a test where the maxPriorityFee is defined to show the resolving is done correctly under the hood?

zhongeric (approved):

test/integ/handlers/check-order-status.test.ts
ConjunctiveNormalForm: good call, added

lib/handlers/check-order-status/service.ts
ConjunctiveNormalForm: found a type from SDK that works!

zhongeric (approved):

PR #483, opened 2024-08-20
fix: bump lambda version 

ConjunctiveNormalForm:
we changed rpc secrets, so need to trigger a response to pick up the new values


codyborn (approved):
LGTM!

PR #487, opened 2024-10-23
build: raise provisioned concurrency to avoid queueing

alanhwu:
There are cases where order webhooks are sent significantly later than their creations.
This causes potential fillers to receive them late and exclusive fillers may miss their exclusivity period, causing a fade.
Orders are sent 1 at a time, and from the logs it seems that lambda instances can have orders queued, which means we may not have enough of them. This problem is more pronounced when a previous execution of a lambda "hangs" from a ECONN or similar error that makes the notification execution 5+ seconds rather than a quick sub-second operation.


zi-yang-uni (approved):

codyborn (approved):
LGTM!

PR #489, opened 2024-11-07
fix: remove kms initialization from injector

zhongeric:


test/unit/handlers/order-notification.test.ts
```diff
@@ -108,7 +108,7 @@ describe('Testing new order Notification handler.', () => {
   it('Testing invalid order with no order hash.', async () => {
     await orderNotificationHandler({ ...MOCK_ORDER, orderHash: undefined })
     expect(logErrorMock).toBeCalledWith(
-      "Error parsing new record to order: Cannot read properties of undefined (reading 'S')",
+      "Error parsing new record to order: Cannot read property 'S' of undefined",
```
zhongeric, line 111: very odd that this failed, but seems reasonable?

zhongeric: ok it has different output on my local vs. in CI

ConjunctiveNormalForm (approved):

PR #494, opened 2024-12-02
feat: add startingBaseFee to dynamoDB

alanhwu:
When inserting an order to the db, we get a `Field 'startingBaseFee' does not have a mapping or alias` error


codyborn (approved):
Looks good, thanks!

PR #496, opened 2024-12-05
Track staleness of v2 orders

codyborn:
Previously we were only tracking staleness when the order is posted.. but it seems like there's a large latency from the time of posting to the time of notification.


lib/handlers/order-notification/handler.ts
```diff
@@ -23,6 +27,22 @@ export class OrderNotificationHandler extends DynamoStreamLambdaHandler<Containe
...
+        // Log the decay start time difference for debugging
+        if (newOrder.orderType == OrderType.Dutch_V2) {
+          const order = CosignedV2DutchOrder.parse(newOrder.encodedOrder, newOrder.chainId)
+          const decayStartTime = order.info.cosignerData.decayStartTime
+          const currentTime = Math.floor(Date.now() / 1000) // Convert to seconds
+          const timeDifference = Number(decayStartTime) - currentTime
+
+          // GPA currentlys sets mainnet decay start to 24 secs into the future
+          if (newOrder.chainId == ChainId.MAINNET && timeDifference > DUTCHV2_ORDER_LATENCY_THRESHOLD_SEC) {
+            const staleOrderMetricName = `NotificationStaleOrder-chain-${newOrder.chainId.toString()}`
```
alanhwu, line 39: Why not hardcode Mainnet in the name since it only runs if `newOrder.chainId == ChainId.MAINNET`?

alanhwu (approved):

PR #499, opened 2024-12-16
Track record staleness

codyborn:
The previous changes didn't improve the order notification staleness. The problem may be upstream so I'd like to see the time difference between the current time and the record's `ApproximateCreationDateTime` (when it's inserted into dynamo).

I also revert the `batchSize` to 1 since that increase didn't help us.


bin/stacks/lambda-stack.ts
```diff
@@ -165,7 +165,7 @@ export class LambdaStack extends cdk.NestedStack {
 
     const notificationConfig = {
       startingPosition: aws_lambda.StartingPosition.TRIM_HORIZON,
-      batchSize: 10,
+      batchSize: 1,
```
ConjunctiveNormalForm, line 168: maybe bump up the provisioned concurrency too, give that with batchSize set to 1 we consistently get 4-6 concurrent lambda executions

ConjunctiveNormalForm (approved):

PR #503, opened 2025-01-06
feat: add fillBlock to table & V3 queries

alanhwu:
- Add `fillBlock` as one of the fields in an order's status
- Make this field show up in queries for DutchV3 orders


codyborn (approved):
Nice! Would this be useful for Priority Orders as well? cc @zhongeric @ConjunctiveNormalForm

PR #508, opened 2025-01-23
Reaper 2.0

codyborn:
I rewrote the broken reaper to properly categorize the orphaned orders. Runs once a day and iterates through block ranges to look for fill events. For the remaining unfilled orders it checks if they have been cancelled (NonceUsed) or expired (Expired) using the quoter validation. 

Note that to make this job stateless, it will only look back 1.2 days (some intentional overlap). However, we currently have over 600 orphaned orders in our database, so I'm using this temporary map to get the oldest block to look back to:
```
export const OLDEST_BLOCK_BY_CHAIN = {
  [ChainId.MAINNET]: 20120259,
  [ChainId.ARBITRUM_ONE]: 253597707,
  [ChainId.BASE]: 22335646,
  [ChainId.UNICHAIN]: 6747397,
}
```


lib/crons/gs-reaper.ts
```diff
@@ -21,37 +25,170 @@ async function main(metrics: MetricsLogger) {
...
+  
+  for (const chainIdKey of Object.keys(OLDEST_BLOCK_BY_CHAIN)) {
+    const chainId = Number(chainIdKey) as keyof typeof OLDEST_BLOCK_BY_CHAIN
+    const provider = providers.get(chainId)
+    if (!provider) {
+      log.error(`No provider found for chainId ${chainId}`)
+      continue
+    }
+
+    const parsedOrders = await getParsedOrders(repo, chainId)
```
ConjunctiveNormalForm, line 60: a comment here explaining what parsedOrders are would be nice

lib/crons/gs-reaper.ts
```diff
@@ -21,37 +25,170 @@ async function main(metrics: MetricsLogger) {
...
+    const orderUpdates = new Map<string, OrderUpdate>()
+
+    // Look through events to find if any of the orders have been filled
+    for (const orderType of Object.keys(REACTOR_ADDRESS_MAPPING[chainId])){
+      const reactorAddress = REACTOR_ADDRESS_MAPPING[chainId][orderType as OrderType]
+      if (!reactorAddress) continue
+      const watcher = new UniswapXEventWatcher(provider, reactorAddress)
+      let lastProcessedBlock = await provider.getBlockNumber()
+      let recentErrors = 0
+      const earliestBlock = OLDEST_BLOCK_BY_CHAIN[chainId]
```
ConjunctiveNormalForm, line 70: so idea is to make a followup pr after this job gets run the first time?

lib/crons/gs-reaper.ts
```diff
@@ -21,37 +25,170 @@ async function main(metrics: MetricsLogger) {
...
+      // const msPerDay = 1000 * 60 * 60 * 24 * 1.2
+      // const blocksPerDay = msPerDay / BLOCK_TIME_MS_BY_CHAIN[chainId]
+      // const earliestBlock = lastProcessedBlock - blocksPerDay
+
+      for (let i = lastProcessedBlock; i > earliestBlock; i -= BLOCK_RANGE) {
+        let attempts = 0
+        while (attempts < CRON_MAX_ATTEMPTS) {
+          try {
+            log.info(`Getting fill events for blocks ${i - BLOCK_RANGE} to ${i}`)
+            const fillEvents = await watcher.getFillEvents(i - BLOCK_RANGE, i)
```
ConjunctiveNormalForm, line 81: there is often a max block window for every eth_getLogs (and similar methods) calls imposed by RPC providers, for example QN says the following in their doc:
`Please be aware that the block range limit for this RPC method is 5 blocks for users on the Free plan and 10,000 blocks for those on paid plans`

so maybe do a loop/promise.all to be safe

lib/crons/gs-reaper.ts
```diff
@@ -21,37 +25,170 @@ async function main(metrics: MetricsLogger) {
...
+            recentErrors = Math.max(0, recentErrors - 1)
+            await Promise.all(fillEvents.map(async (e) => {
+              if (parsedOrders.has(e.orderHash)) {
+                log.info(`Fill event found for order ${e.orderHash}`)
+                // Only get fill info when we know there's a matching event in this
+                // range due to additional RPC calls that are required for fill info
+                const fillInfo = await watcher.getFillInfo(i - BLOCK_RANGE, i)
+                const fillEvent = fillInfo.find((f) => f.orderHash === e.orderHash)
+                if (fillEvent) {
```
ConjunctiveNormalForm, line 90: A lot of this code is from `order-status-check` right? Is it possible to reuse some of the code snippets

or am I mistaken here

ConjunctiveNormalForm (commented):
lgtm! just a few comments

lib/crons/gs-reaper.ts
ConjunctiveNormalForm: ah yeah that should work, my b didn't see the BLOCK_RANGE part

lib/crons/gs-reaper.ts
codyborn: It's tricky because the check order status cron has a lot of logging interspersed which would make it tricky to cleanly encapsulate in a common library.

ConjunctiveNormalForm (approved):

PR #510, opened 2025-01-24
Unimind Algorithm Implementation

alanhwu:
Update parameters based on historical performance for each `pair-chainId`


lib/config/unimind-list.ts
```diff
@@ -0,0 +1,25 @@
+import { Token } from '@uniswap/sdk-core'
+
+const ARBITRUM_CHAIN_ID = 42161
```
codyborn, line 3: this can be pulled from the SDK

lib/crons/unimind-algorithm.ts
```diff
@@ -18,27 +24,163 @@ async function main(metrics: MetricsLogger) {
...
-  repo: UnimindParametersRepository, 
+  unimindParametersRepo: UnimindParametersRepository,
+  ordersRepo: DutchOrdersRepository,
   log: Logger, 
   metrics?: MetricsLogger
 ): Promise<void> {
   const beforeUpdateTime = Date.now()
+  // Query Orders table for latest orders
+  const recentOrders = await getOrdersByTimeRange(ordersRepo, UNIMIND_ALGORITHM_CRON_INTERVAL);
+  log.info(`Found ${recentOrders.length} orders in the last ${UNIMIND_ALGORITHM_CRON_INTERVAL} minutes`)
```
codyborn, line 41: We should prefix all these with "Unimind updateParameters: "

lib/crons/unimind-algorithm.ts
```diff
@@ -18,27 +24,163 @@ async function main(metrics: MetricsLogger) {
...
+): Map<string, number> {
+  const pairCounts = new Map<string, number>();
+
+  for (const order of orders) {
+    let pair = order.pair;
+    // If pair is not already set, create it from input and output tokens
+    if (!pair && order.input && order.outputs) {
+      const inputToken = order.input.token;
+      const outputToken = order.outputs[0].token;
+      pair = `${inputToken}-${outputToken}-${order.chainId}`;
```
codyborn, line 136: should we standardize token addresses to only be lowercase? Should also be applied to `order.pair`

lib/crons/unimind-algorithm.ts
```diff
@@ -18,27 +24,163 @@ async function main(metrics: MetricsLogger) {
...
   const beforeUpdateTime = Date.now()
+  // Query Orders table for latest orders
+  const recentOrders = await getOrdersByTimeRange(ordersRepo, UNIMIND_ALGORITHM_CRON_INTERVAL);
+  log.info(`Found ${recentOrders.length} orders in the last ${UNIMIND_ALGORITHM_CRON_INTERVAL} minutes`)
+  const recentOrderCounts = getOrderCountsByPair(recentOrders);
+  log.info(`Found ${recentOrderCounts.size} unique pairs in the last ${UNIMIND_ALGORITHM_CRON_INTERVAL} minutes`)
+  for (const [pairKey, count] of recentOrderCounts.entries()) {
+    // Get the pair from the unimind parameters table
+    const pairData = await unimindParametersRepo.getByPair(pairKey);
+    if (!pairData) { // We haven't seen this pair before, so it must have received the default parameters
```
codyborn, line 47: nit: can you move this comment to the line above?

lib/crons/unimind-algorithm.ts
```diff
@@ -18,27 +24,163 @@ async function main(metrics: MetricsLogger) {
...
   })
 
-  const repo = DynamoUnimindParametersRepository.create(new DynamoDB.DocumentClient())
-  await updateParameters(repo, log, metrics)
+  const unimindParametersRepo = DynamoUnimindParametersRepository.create(new DynamoDB.DocumentClient())
+  const ordersRepo = DutchOrdersRepository.create(new DynamoDB.DocumentClient()) as DutchOrdersRepository
+  await updateParameters(unimindParametersRepo, ordersRepo, log, metrics)
 }
 
 export async function updateParameters(
```
codyborn, line 32: nit: Can you spell out the algorithm in a natspec comment?

lib/crons/unimind-algorithm.ts
alanhwu: I think the constants we use throughout our codebases are checksummed addresses, and since the pair and input/output token fields are set by URA I think we can just use whatever we get from there

lib/crons/unimind-algorithm.ts
```diff
@@ -18,27 +24,188 @@ async function main(metrics: MetricsLogger) {
...
-  
   const afterUpdateTime = Date.now()
   const updateTime = afterUpdateTime - beforeUpdateTime
   metrics?.putMetric(`unimind-parameters-update-time`, updateTime)
 }
+
+export async function getOrdersByTimeRange(ordersRepo: DutchOrdersRepository, timeRange: number): Promise<UniswapXOrderEntity[]> {
+  // Calculate the timestamp from timeRange minutes ago
+  const timeRangeMs = timeRange * 60 * 1000 // convert minutes to milliseconds
+  const cutoffTime = Math.floor((Date.now() - timeRangeMs) / 1000) // Convert to seconds for DDB
```
codyborn, line 125: nit: why do we convert to ms and then convert to secs?

lib/crons/unimind-algorithm.ts
```diff
@@ -18,27 +24,188 @@ async function main(metrics: MetricsLogger) {
...
+
+  // Query Dutch V3 orders created after the cutoff time
+  const result = await ordersRepo.getOrdersFilteredByType(
+    2000, // reasonable limit for processing
+    {
+      sortKey: SORT_FIELDS.CREATED_AT,
+      sort: `gt(${cutoffTime})`,
+      desc: true,
+      chainId: 42161,
```
codyborn, line 134: nit: pull chainId from sdk

lib/crons/unimind-algorithm.ts
```diff
@@ -18,27 +25,189 @@ async function main(metrics: MetricsLogger) {
...
+interface UnimindStatistics {
+  waitTime: number[];
+  fillStatus: number[];
+  priceImpact: number[];
+}
+
+function getStatistics(orders: DutchV3OrderEntity[]): UnimindStatistics {
+  const waitTime = orders.map(order => {
+    if (!order.fillBlock || !order.cosignerData.decayStartBlock) return -1;
+    return order.fillBlock - order.cosignerData.decayStartBlock;
```
codyborn, line 177: Today we set the `decayStartBlock` to be immediate, so the `fillBlock` is always >= to `decayStartBlock`. However, it's possible for this to change in the future, in which case it would be negative. If that's the case, perhaps the best solution is to treat negative as a valid fill time and return `undefined` for when it's unfilled.

lib/crons/unimind-algorithm.ts
```diff
@@ -18,27 +25,205 @@ async function main(metrics: MetricsLogger) {
...
+function unimindAlgorithm(statistics: UnimindStatistics, pairData: UnimindParameters) {
+  const objective_wait_time = 2;
+  const objective_fill_rate = 0.95;
+  const learning_rate = 2;
+  const auction_duration = 32;
+  const previousParameters = pairData;
+  if (statistics.waitTimes.length === 0 || statistics.fillStatuses.length === 0 || statistics.priceImpacts.length === 0) {
+    return previousParameters;
+  }
+  const average_wait_time = statistics.waitTimes.reduce((a: number, b) => a + (b === undefined ? auction_duration : b), 0) / statistics.waitTimes.length;
```
codyborn, line 216: Do we want to include unfilled orders in these stats? It seems we should remove undefined waitTimes from the denominator. Also we should set negative fill times to be `0` per our slack discussion.

test/unit/util/unimind-algorithm.test.ts
```diff
@@ -0,0 +1,114 @@
+import { getStatistics } from '../../../lib/crons/unimind-algorithm';
+import { DutchV3OrderEntity } from '../../../lib/entities';
+import { ORDER_STATUS } from '../../../lib/entities/Order';
+
+describe('unimind-algorithm', () => {
+  describe('getStatistics', () => {
```
codyborn, line 6: Can you add two more tests for the stats: 
1. the waitTime for unfilled orders is treated as auction duration
2. fills before decay block are considered 0 fill time

codyborn (approved):
Nice! Just left some small comments

PR #513, opened 2025-01-30
feat, test: flag for only posting metadata; default value support

alanhwu:
- Add flag for orders that don't require parameters returned (only seeking to have quote metadata posted)
- Use default values when it's a newly seen pair & append to unimind table
- Tests


lib/handlers/get-unimind/handler.ts
```diff
@@ -8,6 +8,11 @@ import { APIGatewayProxyEvent, APIGatewayProxyResult, Context } from 'aws-lambda
 import { metrics } from '../../util/metrics'
 import { UnimindQueryParams, unimindQueryParamsSchema } from './schema'
 
+const DEFAULT_UNIMIND_PARAMETERS = {
```
codyborn, line 11: Nit: can we move this to a constants file

codyborn (approved):
🌽

PR #518, opened 2025-02-04
feat, test: allow route in metadata to be optional

alanhwu:
- When hitting Unimind endpoint with quote metadata, the route field will be optional


codyborn (approved):
👍

PR #520, opened 2025-02-10
feat, test: store and log classic blockNumber

alanhwu:


zhongeric (approved):
ty! could we add a test showing that blockNumber is persisted and returned please

PR #525, opened 2025-02-24
feat, test: modify classic calldata in order query results with optional executeAddress parameter (V2, V3, Priority)

alanhwu:
Classic calldata is modified to support Artemis as follows:

- Remove Portion command
- Modify Sweep command recipient to Artemis Executor
- Update deadline parameter (if the calldata is `execute()` with deadline)

Queries of Dutch_V2, Dutch_V3, and Priority objects return the route object if it exists

`get-orders` endpoint updated to accept an optional parameter `executeAddress` to have order results returned with calldata modified for use by `executeAddress`

Updated test logging to Bunyan for consistency across codebase


lib/handlers/get-unimind/handler.ts
```diff
@@ -129,4 +142,83 @@ export class GetUnimindHandler extends APIGLambdaHandler<ContainerInjected, Requ
...
+          const sweepInput = inputsArray[sweepIndex]
+          // Decode sweep parameters
+          const [token, , amountMinimum] = defaultAbiCoder.decode(
+              ['address', 'address', 'uint256'],
+              sweepInput
+          )
+          // Encode the parameters with executor address as recipient
+          const modifiedSweepInput = defaultAbiCoder.encode(
+              ['address', 'address', 'uint256'],
+              [token, EXECUTOR_ADDRESS, amountMinimum]
```
codyborn, line 187: 1. It would be great for the route to be usable by any filler
2. We may change the executor address over time and it would be hard to keep this in sync

WDYT about performing this artemisModifyCalldata() in get-orders and adding an optional parameter that we use instead of this value?

lib/handlers/get-orders/schema/GetDutchV2OrderResponse.ts
```diff
@@ -87,4 +89,16 @@ export const GetDutchV2OrderResponseEntryJoi = Joi.object({
   cosignerData: CosignerDataJoi,
   cosignature: Joi.string(),
   createdAt: Joi.number(),
+  route: Joi.object({
```
codyborn, line 92: Nit: I know it's an existing pattern, but maybe we can make this improvement: We could create a `CommonOrderValidationFields` that DutchV2/3 + Priority all share to reduce code duplication.

```
import Joi from 'joi'
import FieldValidator from '../../../util/field-validator'

export const CommonOrderValidationFields = {
  encodedOrder: FieldValidator.isValidEncodedOrder().required(),
  signature: FieldValidator.isValidSignature().required(),
  orderStatus: FieldValidator.isValidOrderStatus().required(),
  orderHash: FieldValidator.isValidOrderHash().required(),
  swapper: FieldValidator.isValidEthAddress().required(),
  chainId: FieldValidator.isValidChainId().required(),
  txHash: FieldValidator.isValidTxHash(),
  quoteId: FieldValidator.isValidQuoteId(),
  requestId: FieldValidator.isValidRequestId(),
  nonce: FieldValidator.isValidNonce(),
  cosignature: Joi.string(),
  createdAt: Joi.number(),
}

export const SettledAmountsValidation = Joi.array().items(
  Joi.object({
    tokenOut: FieldValidator.isValidEthAddress(),
    amountOut: FieldValidator.isValidAmount(),
    tokenIn: FieldValidator.isValidEthAddress(),
    amountIn: FieldValidator.isValidAmount(),
  })
)

export const RouteValidation = Joi.object({
    quote: FieldValidator.isValidAmount(),
    quoteGasAdjusted: FieldValidator.isValidAmount(),
    gasPriceWei: FieldValidator.isValidAmount(),
    gasUseEstimateQuote: FieldValidator.isValidAmount(),
    gasUseEstimate: FieldValidator.isValidAmount(),
    methodParameters: Joi.object({
      calldata: Joi.string(),
      value: Joi.string(),
      to: FieldValidator.isValidEthAddress(),
    }),
  })
```

lib/handlers/get-unimind/handler.ts
```diff
@@ -129,4 +142,83 @@ export class GetUnimindHandler extends APIGLambdaHandler<ContainerInjected, Requ
     const getUnimindRequestByPairMetricName = `GetUnimindRequestPair${pair}`
     metrics.putMetric(getUnimindRequestByPairMetricName, 1, Unit.Count)
   }
-} 
\ No newline at end of file
+}
+
+export function artemisModifyCalldata(calldata: string, log: Logger): string {
+    try {
+      const functionSignatures: Record<string, string> = {
```
codyborn, line 149: nit: this can be moved to a constants file

lib/handlers/get-unimind/handler.ts
```diff
@@ -129,4 +142,83 @@ export class GetUnimindHandler extends APIGLambdaHandler<ContainerInjected, Requ
...
+        "24856bc3": "function execute(bytes commands, bytes[] inputs)",
+        "3593564c": "function execute(bytes commands, bytes[] inputs, uint256 deadline)"
+      };
+
+      const functionSelector = calldata.slice(2, 10)
```
codyborn, line 154: Can they be derived in the code to make it explicit? Something like:
```suggestion
      const prefix = "0x"
      const functionSelectorBytes = 4
      const charsPerByte = 2
      const functionSelector = calldata.slice(prefix.length, prefix.length + functionSelectorBytes * charsPerByte)
```

lib/handlers/get-unimind/handler.ts
```diff
@@ -129,4 +142,83 @@ export class GetUnimindHandler extends APIGLambdaHandler<ContainerInjected, Requ
     const getUnimindRequestByPairMetricName = `GetUnimindRequestPair${pair}`
     metrics.putMetric(getUnimindRequestByPairMetricName, 1, Unit.Count)
   }
-} 
\ No newline at end of file
+}
+
+export function artemisModifyCalldata(calldata: string, log: Logger): string {
```
codyborn, line 147: Is there a way we can reduce the size of the function and make it more abstract? Eg. create a UniversalRouterRoute class that takes in the calldata in the constructor and we can perform read/write operations on it

lib/handlers/get-unimind/handler.ts
```diff
@@ -129,4 +142,83 @@ export class GetUnimindHandler extends APIGLambdaHandler<ContainerInjected, Requ
...
+              [token, EXECUTOR_ADDRESS, amountMinimum]
+          )
+          inputsArray[sweepIndex] = modifiedSweepInput
+      }
+      
+      // Re-encode the complete calldata
+      let modifiedCalldata;
+      if (functionSelector == "24856bc3") {
+          modifiedCalldata = iface.encodeFunctionData('execute', [commandArray, inputsArray])
+      } else if (functionSelector == "3593564c") {
```
codyborn, line 196: nit: these should all be constants
`3593564c`, `24856bc3`, `execute`

codyborn (commented):
Looks great! Left some questions and suggestions.

lib/handlers/get-unimind/handler.ts
alanhwu: done! Added an optional parameter called `executeAddress`

lib/handlers/get-unimind/handler.ts
alanhwu: Created a `UniversalRouterCalldata` class that abstracts these operations away!

lib/util/UniversalRouterCalldata.ts
```diff
@@ -0,0 +1,137 @@
...
+
+export class UniversalRouterCalldata {
+	private iface: Interface;
+	private signature: string;
+	private functionSelector: string;
+	private commandArray: CommandType[];
+	private inputsArray: any[];
+	private log: Logger;
+
+	constructor(calldata: string, log: Logger) {
```
codyborn, line 14: nit: looks like there's some extra whitespace here

lib/util/UniversalRouterCalldata.ts
```diff
@@ -0,0 +1,137 @@
...
+			calldata
+		});
+		return "";
+	}
+}
+
+function getCommands(commands: string): CommandType[] {
+	const commandTypes = []
+
+	for (let i = 2; i < commands.length; i += 2) {
```
codyborn, line 131: i = 2 because the prefix "0x", right? Could you add this as a comment?

lib/util/UniversalRouterCalldata.ts
```diff
@@ -0,0 +1,137 @@
...
+		const byte = commands.substring(i, i + 2)
```
codyborn, line 132: If commands.length is odd, this would throw an out of bounds exception. We can set the loop condition to `i + 2 < commands.length` to prevent this.

lib/util/UniversalRouterCalldata.ts
```diff
@@ -0,0 +1,137 @@
...
+				['address', 'address', 'uint256'],
+				sweepInput
+			);
+			// Encode the parameters with new recipient address
+			const modifiedSweepInput = defaultAbiCoder.encode(
+				['address', 'address', 'uint256'],
+				[token, recipient, amountMinimum]
+			);
+			this.inputsArray[sweepIndex] = modifiedSweepInput;
+			this.log.info('Modified SWEEP command recipient');
```
codyborn, line 82: nit: we should remove these log statements before merging

codyborn (approved):
Looks good! Left some small comments

lib/util/UniversalRouterCalldata.ts
alanhwu: Addressed this with an even validation before the loop

PR #532, opened 2025-03-05
Add disabled API keys

codyborn:
Adding API keys for the following endpoints:
- GET /unimind
- POST /orders
- POST /limit-orders

All are currently disabled. Will enable them once we have all callers integrated with new keys (Trading API, URA, CLIs, and E2E tests)
image


alanhwu (approved):
Makes sense! Will need to have a dev key for `frequent-swappers` as well when we roll this out

ConjunctiveNormalForm (commented):
lgtm!

PR #533, opened 2025-03-10
feat, test: add pair as a GSI in Orders table

alanhwu:
Unimind update algorithm needs to perform queries on `pair` in the `Orders` table
1. Adds `pair` as a GSI to enable queries on the `pair` attribute
2. Updates the `get-orders` endpoint to utilize this GSI to allow API users to also query for orders based on `pair`

### DIFF
--- a/test/unit/builders/QueryParamsBuilder.ts
+++ b/test/unit/builders/QueryParamsBuilder.ts
@@ -24,6 +24,11 @@ export class QueryParamsBuilder {
     return this
   }
 
+  withPair(value?: string) {
+    this.params.pair = value || 'ETH-USDC-1'
+    return this
+  }
+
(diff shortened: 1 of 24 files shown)


test/unit/builders/QueryParamsBuilder.ts
```diff
@@ -24,6 +24,11 @@ export class QueryParamsBuilder {
     return this
   }
 
+  withPair(value?: string) {
+    this.params.pair = value || 'ETH-USDC-1'
```
codyborn, line 28: these should be addresses, not symbols, right?

codyborn (approved):
Looks good! Just a reminder to update unimind.ts in URA

PR #538, opened 2025-03-20
feat: usedUnimind flag and analytics

alanhwu:
Add a `usedUnimind` flag for quotes that were given Unimind parameters
Orders ultimately submitted from those quotes will be recorded and logged with the boolean `usedUnimind`

### DIFF
--- a/lib/services/analytics-service.ts
+++ b/lib/services/analytics-service.ts
@@ -37,6 +37,7 @@ export class AnalyticsService implements AnalyticsServiceInterface {
       orderType: orderType,
       blockNumber: order?.blockNumber,
       route: JSON.stringify(order?.route),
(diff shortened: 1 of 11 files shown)


lib/services/analytics-service.ts
```diff
@@ -37,6 +37,7 @@ export class AnalyticsService implements AnalyticsServiceInterface {
       orderType: orderType,
       blockNumber: order?.blockNumber,
       route: JSON.stringify(order?.route),
+      usedUnimind: order?.usedUnimind,
```
codyborn, line 40: If order is undefined, then `useUnimind` will be set to `undefined`, not `false`. That doesn't seem to be what's intended since usedUnimind is set to required in lib/repositories/quote-metadata-repository.ts

PR #539, opened 2025-03-21
feat, test: move Unimind intrinsic values into a map for extensibility

alanhwu:
- We would like to store intrinsic values beyond pi & tau in the future
- Instead of having these as fields in the `UnimindParameters` table, we will consolidate them into a `intrinsicValues` field in the form of a Map
- This will make future changes to our choice of intrinsic values not require database-level changes and only in operations that RW from the table


lib/crons/unimind-algorithm.ts
```diff
@@ -231,8 +236,19 @@ export function unimindAlgorithm(statistics: UnimindStatistics, pairData: Unimin
...
+  //return a record of pi and tau
   return {
-    pi,
-    tau,
+    pi: pi,
+    tau: tau,
   };
+}
+
+export function validateParameters(parameters: UnimindParameters): boolean {
```
codyborn, line 246: It might be better to store the values as a serialized json object. It would allow us to store different value types (e.g. BigNumber) and it makes versioning easier to reason about.

codyborn (approved):

PR #543, opened 2025-03-27
feat, test: support modifying calldata for swaps unwrapping WETH

alanhwu:
- Observed some swaps using the `UNWRAP_WETH` command from `UniversalRouter`
- It is being used in lieu of `SWEEP`, so the `recipient` field must be updated the same way we update the `SWEEP` recipient for most swaps


lib/util/UniversalRouterCalldata.ts
```diff
@@ -81,6 +81,25 @@ export class UniversalRouterCalldata {
     return this;
   }
 
+  public modifyUnwrapRecipient(recipient: string): UniversalRouterCalldata {
+    const unwrapIndex = this.commandArray.findIndex(command => command == CommandType.UNWRAP_WETH);
+    if (unwrapIndex !== -1) {
+      const unwrapInput = this.inputsArray[unwrapIndex];
+      // Decode unwrap parameters
+      const [, amountMin] = defaultAbiCoder.decode(
+        ['address', 'uint256'],
```
codyborn, line 90: nit: we can add these as constants somewhere so a typo doesn't get us at some point

codyborn (approved):
Nice!

PR #545, opened 2025-03-30
feat, test: price impact based algorithm

alanhwu:
Maintained variable and function names to match Python implementation for ease of comparison


lib/util/constants.ts
```diff
@@ -26,11 +26,13 @@ export const REAPER_RANGES_PER_RUN = 10
 export const DYNAMO_BATCH_WRITE_MAX = 25
 
 export const DEFAULT_UNIMIND_PARAMETERS = JSON.stringify({
-  pi: 5,
-  tau: 5
+  lambda1: 0,
+  lambda2: 5,
+  Sigma: 0.0003
```
codyborn, line 31: Is there a reason Sigma should be capitalized?

lib/util/unimind.ts
```diff
@@ -1,5 +1,204 @@
...
+  
+  // Extract current parameters
+  const lambda1 = previousParameters.lambda1;
+  const lambda2 = previousParameters.lambda2;
+  const Sigma = previousParameters.Sigma;
+  
+  // Create array of valid data points (the calculation is based on filled orders)
+  const validDataPoints: { waitTime: number, priceImpact: number, fillStatus: number }[] = [];
+  for (let i = 0; i < waitTimes.length; i++) {
+    if (waitTimes[i] !== undefined && priceImpacts[i] !== undefined) {
```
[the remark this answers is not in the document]
codyborn: Oh, nvm I see: `unfilled orders are being ignored`

lib/util/unimind.ts
```diff
@@ -1,5 +1,204 @@
...
+  
+  // Log stats about what data we're using
+  log.info({
+    totalOrders: waitTimes.length,
+    usedForOptimization: validDataPoints.length,
+    ignoredOrders: waitTimes.length - validDataPoints.length
+  }, 'Unimind algorithm data usage - NOTE: unfilled orders are being ignored');
+  
+  // Calculate gradients for each filled order
+  const gradients = validDataPoints.map(({ waitTime, priceImpact }) => {
```
codyborn, line 68: oh boy.. these variable names 😅 can you wrap this in a separate function and we can independently unit test?

lib/util/unimind.ts
```diff
@@ -1,5 +1,204 @@
...
+  // Update parameters using gradient descent
+  const lambda1_updated = lambda1 - lambda1_learning_rate * d_J_d_lambda1_avg;
+  const lambda2_updated = lambda2 - lambda2_learning_rate * d_J_d_lambda2_avg;
+
+  // Update Sigma with fill rate
+  // Calculate average fill rate for the batch
+  const fill_rate = statistics.fillStatuses.reduce((sum, status) => sum + status, 0) / statistics.fillStatuses.length;
+
+  const d_J_d_FR = 2 * beta * (fill_rate - target_fill_rate);
+  const d_FR_d_Sigma = Math.log(0.00001); // Assume constant for now
```
codyborn, line 113: nit: declare as constant

lib/util/unimind.ts
```diff
@@ -1,5 +1,204 @@
...
+  const price_impact_of_amm = extrinsicValues.priceImpact
+  const price_impact_filler = compute_price_impact_filler(price_impact_of_amm, intrinsicValues)
+  return (price_impact_of_amm- price_impact_filler) / (1 - price_impact_of_amm)
+}
+
+function compute_price_impact_filler(price_impact_of_amm: number, intrinsicValues: any) {
+  const lambda1 = intrinsicValues.lambda1
+  const lambda2 = intrinsicValues.lambda2
+
+  const Lambda2 = (2 / (1 + Math.exp(-lambda2))) - 1
```
codyborn, line 154: this is hard for me to reason about, can you add a comment to explain what's going on?

lib/util/unimind.ts
codyborn: Did you explore using existing math/ML libraries to do some of the heavy lifting? Eg. tensorFlow, optimization-js

codyborn: If those are overkill, we can define our own math lib just to abstract the math operations here

lib/util/unimind.ts
```diff
@@ -1,5 +1,204 @@
...
+    lambda1: lambda1_updated,
+    lambda2: lambda2_updated,
+    Sigma: Sigma_updated
+  };
+}
+
+export function computePi(intrinsicValues: any, extrinsicValues: QuoteMetadata) {
+  const price_impact_of_amm = extrinsicValues.priceImpact
+  const price_impact_filler = compute_price_impact_filler(price_impact_of_amm, intrinsicValues)
+  return (price_impact_of_amm- price_impact_filler) / (1 - price_impact_of_amm)
```
codyborn, line 147: Should check if `price_impact_of_amm === 1` to prevent divide by 0

lib/util/unimind.ts
```diff
@@ -1,5 +1,204 @@
...
\ No newline at end of file
+}
+
+/**
+ * @notice Adjusts Unimind parameters (intrinsic values) based on historical order statistics
+ * @param statistics Aggregated order data containing arrays of wait times, fill statuses, and price impacts
+ * @param pairData Previous parameters intrinsic values stored for the pair
+ * @return Updated pi and tau parameters
+ */
+export function unimindAlgorithm(statistics: UnimindStatistics, pairData: UnimindParameters, log: Logger) {
```
codyborn, line 19: maybe a question for Brad.. are there any assumptions about the statistics (e.g. valid ranges) that we should check for?

lib/util/unimind.ts
```diff
@@ -1,5 +1,204 @@
...
+  // Extract current parameters
+  const lambda1 = previousParameters.lambda1;
+  const lambda2 = previousParameters.lambda2;
+  const Sigma = previousParameters.Sigma;
+  
+  // Create array of valid data points (the calculation is based on filled orders)
+  const validDataPoints: { waitTime: number, priceImpact: number, fillStatus: number }[] = [];
+  for (let i = 0; i < waitTimes.length; i++) {
+    if (waitTimes[i] !== undefined && priceImpacts[i] !== undefined) {
+      validDataPoints.push({
```
codyborn, line 52: Is there an upper bound on max number of orders to process? Thinking about memory consumption here if we get a burst of spam traffic for example.

lib/util/constants.ts
alanhwu: We express it as a capital sigma which is a different symbol from the lowercase sigma so it was like this in the Python code

codyborn (approved):

PR #549, opened 2025-04-16
fix, test: price impact strategy should use price impacts as fractions

alanhwu:
- The algorithm expects the range of `priceImpact` to be `[0,1]` but in the rest of our systems we express it as percentage points `[0,100]`
- This PR normalizes the values back to fractions only in the Price Impact strategy implementation to abstract this difference

### DIFF
--- a/.github/workflows/semgrep.yml
+++ b/.github/workflows/semgrep.yml
@@ -11,7 +11,7 @@ on:
 jobs:
   semgrep:
     name: semgrep/ci
-    runs-on: uubuntu-latest
+    runs-on: ubuntu-latest
     env:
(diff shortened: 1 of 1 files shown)


.github/workflows/semgrep.yml
```diff
@@ -11,7 +11,7 @@ on:
 jobs:
   semgrep:
     name: semgrep/ci
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
```
codyborn, line 14: what is this change for?

alanhwu: 20.04 got deprecated today so it was blocking the CI

PR #550, opened 2025-04-18
feat: enable unimind 1% flight with token list

alanhwu:
- Enables Unimind for 1% of public addresses
- Decreases update threshold from 50 -> 25 orders
- Uses token list supported by Unimind for initial order flow
- Log and maintain metric for public addresses that used Unimind for debugging purposes


lib/util/unimind.ts
```diff
@@ -3,9 +3,28 @@ import { UnimindParameters } from "../repositories/unimind-parameters-repository
...
 
 export function unimindAddressFilter(address: string) {
-  return address.toLowerCase() === UNIMIND_DEV_SWAPPER_ADDRESS.toLowerCase()
+  // Always include the dev swapper address
+  if (address.toLowerCase() === UNIMIND_DEV_SWAPPER_ADDRESS.toLowerCase()) {
+    return true;
+  }
+  
+  // Sample 1% of other addresses by using the last 4 hex characters
+  const lastFourChars = address.slice(-4);
```
codyborn, line 15: Prob not a big deal, but it might not be a random 1% if some people have a vanity address with 0s. We could hash the address before converting to an int.

lib/util/unimind.ts
```diff
@@ -3,9 +3,28 @@ import { UnimindParameters } from "../repositories/unimind-parameters-repository
...
+  const value = parseInt(lastFourChars, 16);
+  return value % 100 === 0; // 1% chance
```
codyborn, line 17: To make this easier to ramp up in the future, we can use:
```
const samplePercent = 1;
return (value % 100) + 1 <= samplePercent;

File: lib/util/unimind.ts [hunk 3]
```diff
@@ -3,9 +3,28 @@ import { UnimindParameters } from "../repositories/unimind-parameters-repository
 import { UnimindStatistics } from "../crons/unimind-algorithm";
 import { default as Logger } from 'bunyan'
 import { QuoteMetadata } from "../repositories/quote-metadata-repository";
+import { UNIMIND_LIST } from "../config/unimind-list";
 
 export function unimindAddressFilter(address: string) {
```
codyborn, line 8: can you add some tests for the sampling?

test/unit/util/unimind-algorithm.test.ts
```diff
@@ -199,4 +200,38 @@ describe('unimind-algorithm', () => {
...
+    });
+
+    // Count how many addresses pass the filter
+    const passedCount = addresses.filter(addr => 
+      unimindAddressFilter(addr)
+    ).length;
+
+    // Check that the percentage is approximately UNIMIND_SAMPLE_PERCENT%
+    const percentage = (passedCount / sampleSize) * 100;
+    expect(percentage).toBeGreaterThanOrEqual(UNIMIND_SAMPLE_PERCENT - 0.5);
```
codyborn, line 234: we might want to allow more wiggle room to not introduce false positives

codyborn (approved):

PR #555, opened 2025-05-01
feat, test: versioning, expand token list

alanhwu:
- Expand token list
- Support `UnimindParameters` versioning. If fetched parameters from the DB are from an old version of the algorithm, Unimind should overwrite them or act using default parameters instead


lib/repositories/unimind-parameters-repository.ts
```diff
@@ -36,6 +37,7 @@ export class DynamoUnimindParametersRepository implements UnimindParametersRepos
         pair: { partitionKey: true, type: DYNAMODB_TYPES.STRING },
         intrinsicValues: { type: DYNAMODB_TYPES.STRING, required: true },
         count: { type: DYNAMODB_TYPES.NUMBER, required: true },
+        version: { type: DYNAMODB_TYPES.NUMBER, required: true },
```
codyborn, line 40: Can you make this required if the old records don't have it?

codyborn (approved):

PR #556, opened 2025-05-01
feat, test: update V4_SWAP action recipient to Executor address

alanhwu:
- When modifying calldata for use with an executor, we also need to modify the recipient of the `TAKE` action when a V4 Swap is performed


lib/util/UniversalRouterCalldata.ts
```diff
@@ -107,6 +108,56 @@ export class UniversalRouterCalldata {
...
+      (command) => command === CommandType.V4_SWAP
+    )
+
+    if (v4SwapIndex === -1) return this
+
+    const v4Input = this.inputsArray[v4SwapIndex] as string
+
+    // Decode wrapper structure to (actionsHex, params[])
+    const [actionsHex, paramsArray]: [string, string[]] = defaultAbiCoder.decode(
+      ['bytes', 'bytes[]'],
```
codyborn, line 123: nit: add to constants

lib/util/UniversalRouterCalldata.ts
```diff
@@ -107,6 +108,56 @@ export class UniversalRouterCalldata {
...
+
+    const bytesWithout0x = actionsHex.startsWith('0x') ? actionsHex.slice(2) : actionsHex
+    const updatedParams = [...paramsArray]
+
+    // Go through actions to rewrite recipient
+    for (let i = 0; i < bytesWithout0x.length; i += 2) {
+      const actionByte = parseInt(bytesWithout0x.slice(i, i + 2), 16) as Actions
+
+      if (actionByte === Actions.TAKE) {
+        const paramIndex = i / 2
```
codyborn, line 135: Not sure I follow why the param index is half the current index. Could you add a comment explaining how this always is the case?

lib/util/UniversalRouterCalldata.ts
```diff
@@ -139,6 +190,7 @@ export function artemisModifyCalldata(calldata: string, log: Logger, executeAddr
       .removePayPortionCommand()
       .modifySweepRecipient(executeAddress)
       .modifyUnwrapRecipient(executeAddress)
+      .modifyV4SwapRecipient(executeAddress)
       .encode();
```
codyborn, line 194: Can you add the check here to search the final encoded value for any of the original recipient addresses and if detected, throw out the result and log the calldata in an error log?

codyborn (approved):

lib/util/UniversalRouterCalldata.ts
codyborn: Added

PR #557, opened 2025-05-02
feat: log fill block for DutchV3 orders

alanhwu:
For DutchV3 orders, we can calculate an exact fill time in units of blocks. Logging it in `FillEventLogger` will propagate it to the Redshift S3 so that we can ingest it into BigQuery to add to our dashboards.

### DIFF
--- a/lib/handlers/check-order-status/service.ts
+++ b/lib/handlers/check-order-status/service.ts
@@ -20,8 +20,9 @@ import { ChainId } from '../../util/chain'
 import { metrics } from '../../util/metrics'
 import { SfnStateInputOutput } from '../base'
 import { FillEventLogger } from './fill-event-logger'
-import { getSettledAmounts, IS_TERMINAL_STATE } from './util'
+import { getSettledAmounts, IS_TERMINAL_STATE, timestampToBlockNumber } from './util'
 import { parseOrder } from '../OrderParser'
+import { PRIORITY_ORDER_TARGET_BLOCK_BUFFER } from '../constants'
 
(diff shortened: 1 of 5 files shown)


lib/handlers/check-order-status/service.ts
```diff
@@ -90,7 +90,10 @@ export class CheckOrderStatusService {
     )
 
     const fromBlock = !startingBlockNumber ? curBlockNumber - this.fillEventBlockLookback(chainId) : startingBlockNumber
-
+    let fillTimeBlocks = startingBlockNumber ? curBlockNumber - startingBlockNumber : undefined // Approximate for non-DutchV3 orders
```
codyborn, line 93: This runs as a step function on a certain interval. It may be that the fill block and the current block are different. The fill block could be anything between current block and fillEventBlockLookback.

lib/handlers/check-order-status/service.ts
```diff
@@ -119,6 +119,12 @@ export class CheckOrderStatusService {
             provider.getTransaction(fillEvent.txHash),
             provider.getBlock(fillEvent.blockNumber),
           ])
+
+          let fillTimeBlocks = startingBlockNumber ? block.number - startingBlockNumber : undefined // Approximate for non-DutchV3 orders
```
codyborn, line 123: `startingBlockNumber` is just the start of the search filter, right? It doesn't seem like this would be a helpful representation of fill time in blocks. For Dutch V2 & v3 it should be `decayStartBlock` like you have below. For Priority Orders, it should be the target block.

codyborn (approved):

PR #572, opened 2025-06-04
feat: don't retry failed webhooks; add notifiedAt field; abstract logging; measure endpoint fetching

alanhwu:


zodahu:
Thanks for this PR! We're a new filler on Base encountering webhook delays of about 5–30 seconds after the `auctionTargetBlock`. This PR will be really helpful for debugging that issue.

codyborn (approved):

PR #583, opened 2025-06-12
fix: allow 0 priceImpact

alanhwu:
After the price impact calculation fixes, sometimes price impact can be equal to 0
This would short-circuit a check that was intended to see if the field was accessible

### DIFF
--- a/lib/crons/unimind-algorithm.ts
+++ b/lib/crons/unimind-algorithm.ts
@@ -193,14 +193,14 @@ export function getStatistics(orders: DutchV3OrderEntity[], log: Logger): Unimin
   const priceImpacts: number[] = [];
 
   for (const order of orders) {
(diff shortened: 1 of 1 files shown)


lib/crons/unimind-algorithm.ts
```diff
@@ -193,14 +193,14 @@ export function getStatistics(orders: DutchV3OrderEntity[], log: Logger): Unimin
   const priceImpacts: number[] = [];
 
   for (const order of orders) {
-    if (order.fillBlock && order.cosignerData?.decayStartBlock && order.priceImpact && 
+    if (order.fillBlock && order.cosignerData?.decayStartBlock && order.priceImpact != null &&
```
codyborn, line 196: Should this be null or undefined (or both)?

codyborn: Sweet, I didn't know that!

codyborn (approved):

PR #589, opened 2025-07-07
feat: add dashboard panel for proportion of Unimind orders with negative pi parameter

alanhwu:
- Fix typo in the requests/responses panel
- Add panel for the % of the Unimind orders that are suggested an auction start price lower than the AMM
- Add panel for the percentiles of pi values
- Add panel for the histogram of pi values (bucketed by 1bps)

### DIFF
--- a/bin/stacks/dashboard-stack.ts
+++ b/bin/stacks/dashboard-stack.ts
@@ -592,6 +592,32 @@ export class DashboardStack extends cdk.NestedStack {
               title: 'Unimind Calculation Times (ms)',
             },
           },
+          {
+            height: 6,
+            width: 12,
+            y: 6,
+            x: 0,
+            type: 'metric',
+            properties: {
+              metrics: [
+                [{ expression: '(negativePi/successfulRequests) * 100', label: 'Negative Pi Percentage', id: 'e1' }],
+                ['Uniswap', 'UnimindNegativePi', 'Service', 'UniswapXService', { id: 'negativePi', visible: false }],
+                ['Uniswap', 'GetUnimindStatus2XX', 'Service', 'UniswapXService', { id: 'successfulRequests', visible: false }],
+              ],
+              view: 'timeSeries',
+              stacked: false,
+              region,
+              stat: 'Sum',
+              period: 300,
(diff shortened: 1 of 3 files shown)


lib/handlers/get-unimind/handler.ts
```diff
@@ -91,6 +91,16 @@ export class GetUnimindHandler extends APIGLambdaHandler<ContainerInjected, Requ
       const calculateTime = afterCalculateTime - beforeCalculateTime
       metrics.putMetric(`final-parameters-calculation-time`, calculateTime)
 
+      // Track negative and positive pi values
+      if (parameters.pi < 0) {
```
codyborn, line 95: WDYT about using a histogram instead of a binary metric? We would log the actual pi value and then plot on a histogram.
https://docs.aws.amazon.com/quicksight/latest/user/histogram-charts.html

alanhwu: added percentile & histogram

bin/stacks/dashboard-stack.ts
```diff
@@ -589,7 +597,83 @@ export class DashboardStack extends cdk.NestedStack {
...
+            },
+          },
+          {
+            height: 6,
+            width: 24,
+            y: 12,
+            x: 0,
+            type: 'log',
+            properties: {
+              query: `SOURCE '/aws/lambda/${getUnimindLambdaName}' | fields floor(pi) as bucket | filter eventType = "UnimindPiCalculated" | filter bucket >= -15 and bucket <= 15 | stats count() by bucket | sort bucket asc`,
```
codyborn, line 672: Maybe we can include the outliers in the min and max bucket with something like this:
`SOURCE '/aws/lambda/${getUnimindLambdaName}' | fields max(min(floor(pi), 15),-15) as bucket | filter eventType = "UnimindPiCalculated" | stats count() by bucket | sort bucket asc`

codyborn: Also, can `UnimindNegativePi` and `UnimindPositivePi` be expressed similarly? So we would only need to record `UnimindPiCalculated` in the code.

codyborn (approved):

PR #594, opened 2025-08-08
fix: permissioned token transfer check

ConjunctiveNormalForm:
for permissioned tokens, don't call the onchainValidator

### DIFF
--- a/lib/services/UniswapXOrderService.ts
+++ b/lib/services/UniswapXOrderService.ts
@@ -111,9 +111,6 @@ export class UniswapXOrderService {
       throw new OrderValidationFailedError(offChainValidationResult.errorString)
     }
 
-    // Still considered valid
-    if (order instanceof CosignedPriorityOrder && onChainValidationResult == OrderValidation.OrderNotFillableYet) return
-
     if (PermissionedTokenValidator.isPermissionedToken(order.info.input.token, chainId)) {
       const provider = this.providerMap.get(chainId)
       if (!provider) {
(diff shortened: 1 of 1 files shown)


lib/services/UniswapXOrderService.ts
```diff
@@ -116,11 +117,39 @@ export class UniswapXOrderService {
...
+      // TODO: excluded v3 orders because harder to handle trade type etc
+      if (order instanceof PriorityOrder || order instanceof DutchV3Order) {
+        throw new OrderValidationFailedError("Permissioned tokens shouldn't work with priority/v3 dutch orders")
+      }
+      const permissionedOrder = order as DutchOrder | CosignedV2DutchOrder
+      const exclusiveFiller =
+        permissionedOrder instanceof DutchOrder
+          ? permissionedOrder.info.exclusiveFiller
+          : permissionedOrder.info.cosignerData.exclusiveFiller
+      const preTransferCheckResult = await PermissionedTokenValidator.preTransferCheck(
```
marktoda, line 135: assume this is only checking whether the swapper and filler are allowlisted? wonder if worth doing any other validation

lib/services/UniswapXOrderService.ts
```diff
@@ -116,11 +117,39 @@ export class UniswapXOrderService {
     // Still considered valid
     if (order instanceof CosignedPriorityOrder && onChainValidationResult == OrderValidation.OrderNotFillableYet) return
 
-    if (onChainValidationResult !== OrderValidation.OK) {
-      const failureReason = OrderValidation[onChainValidationResult]
-      throw new OrderValidationFailedError(`Onchain validation failed: ${failureReason}`)
+    if (PermissionedTokenValidator.isPermissionedToken(order.info.input.token, chainId)) {
```
marktoda, line 120: isnt the onChainValidator.validate() call above still going to fail before we get here?

lib/services/UniswapXOrderService.ts
ConjunctiveNormalForm: I feel like we should do them in PermissionedTokenValidator too, and can be worked on later

PR #595, opened 2025-08-08
feat, test: immediately notify exclusive filler

alanhwu:
Dynamo Streams aren't very low-latency, so the notification lambda is occasionally not triggered fast enough. This can cause exclusive fillers to be late on fills on Mainnet. This change ensures that exclusive fillers are notified as soon as an order is created instead of waiting for it to be picked up by Dynamo Streams.

### DIFF
--- a/lib/handlers/order-notification/handler.ts
+++ b/lib/handlers/order-notification/handler.ts
@@ -41,54 +41,7 @@ export class OrderNotificationHandler extends DynamoStreamLambdaHandler<Containe
 
         log.info({ order: newOrder, registeredEndpoints }, 'Sending order to registered webhooks.')
 
-        // Randomize the order to prevent any filler from having a consistent advantage
-        const shuffledEndpoints = [...registeredEndpoints].sort(() => Math.random())
-        const requests: Promise<AxiosResponse>[] = shuffledEndpoints.map((endpoint) =>
-          axios.post(
-            endpoint.url,
-            {
-              orderHash: newOrder.orderHash,
-              createdAt: newOrder.createdAt,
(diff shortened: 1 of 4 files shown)


lib/services/UniswapXOrderService.ts
```diff
@@ -94,6 +97,35 @@ export class UniswapXOrderService {
     const realOrderType = order.orderType
     await this.logOrderCreatedEvent(orderEntity, realOrderType)
 
+    // Send immediate notification to exclusive filler if present
+    if (this.webhookProvider && orderEntity.filler) {
```
codyborn, line 101: IIRC in the case of non-exclusive filler, the `orderEntity.filler` would be the zero address (0x000...).

lib/handlers/order-notification/handler.ts
```diff
@@ -161,3 +114,150 @@ export class OrderNotificationHandler extends DynamoStreamLambdaHandler<Containe
...
   }
 }
+
+/**
+ * Send webhook notifications to all provided endpoints
+ * Extracted from OrderNotificationHandler for reuse in immediate notifications
+ */
+export async function sendWebhookNotifications(
+  endpoints: Array<{ url: string; headers?: { [key: string]: string } }>,
+  order: {
```
codyborn, line 124: Can we define an order type that both of these endpoints can use? Perhaps it can use a Partial<SharedXOrderEntity>?

lib/services/UniswapXOrderService.ts
alanhwu: Good catch!! Made a helper function to check it

PR #605, opened 2025-08-26
Lower target block buffer on Base + Unichain

codyborn:
Decrease start time by 2 seconds on both Base (from 6s) and Unichain (from 5s)


alanhwu (approved):

codyborn:
Creating a new PR that will keep timing more consistent

PR #609, opened 2025-08-29
Consistent priority auction start time

codyborn:
Lowering auction start buffer for Priority Orders
Additionally, adding an extra block IF the current block is too stale (block time * 75%). This ensures the buffer is always at least 1.25 + PRIORITY_ORDER_TARGET_BLOCK_BUFFER.

### DIFF
--- a/lib/models/PriorityOrder.ts
+++ b/lib/models/PriorityOrder.ts
@@ -117,10 +117,9 @@ export class PriorityOrder extends Order {
     // This keeps the time window more consistent for fillers
     const blockTimeSeconds = AVERAGE_BLOCK_TIME(this.chainId as ChainId)
     const timeDifference = Math.abs(currentTime - block.timestamp)
-    const halfBlockTime = blockTimeSeconds * 0.75
     
     // If the difference is more than 75% of the block time, add an extra block
-    const extraBlock = timeDifference > halfBlockTime ? 1 : 0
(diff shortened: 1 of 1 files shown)


lib/models/PriorityOrder.ts
```diff
@@ -109,9 +110,22 @@ export class PriorityOrder extends Order {
...
   public async reparameterizeAndCosign(provider: StaticJsonRpcProvider, cosigner: KmsSigner): Promise<this> {
-    const currentBlock = await provider.getBlockNumber()
+    const block = await provider.getBlock("latest")
+    const currentTime = Math.floor(Date.now() / 1000) // Current time in seconds
+    
+    // Calculate if we need to add an extra block based on timestamp difference
+    // This keeps the time window more consistent for fillers
+    const blockTimeSeconds = AVERAGE_BLOCK_TIME(this.chainId as ChainId)
+    const timeDifference = Math.abs(currentTime - block.timestamp)
+    const halfBlockTime = blockTimeSeconds * 0.75
```
ConjunctiveNormalForm, line 120: nit: not half block time though?

codyborn: If it's less than .75 the way through, the most that will be extra is 1 - 1.25 blocks. If it's more than .75 through, we add a block so the extra will be 1.25 - 1 blocks.

If we used .50, then it'd be 1 - .5 vs 1.5 - 1.

PR #614, opened 2025-09-08
feat, test: A/B test Unimind on per-trade basis

alanhwu:


codyborn (approved):
Looks good. Just a reminder to change UNIMIND_ADDRESS_SAMPLE_PERCENT to 66% so when we ramp up X traffic to 75%, we get 50% (66% of 75%) going to Unimind and 25% as the control.

codyborn (approved):

PR #616, opened 2025-09-10
feat: Unimind response and update data logging

alanhwu:
Lays the foundation for getting Unimind responses & updates into GPA and BigQuery


lib/services/analytics-service.ts
```diff
@@ -10,6 +10,38 @@ export interface AnalyticsServiceInterface {
...
+  route?: any
+  amountIn?: string
+  amountOut?: string
+  chainId?: number
+}
+
+// Parameters for Unimind parameter update analytics
+export interface UnimindParameterUpdateParams {
+  pair: string
+  updateType: 'new_pair' | 'algorithm_update' | 'threshold_reached'
```
codyborn, line 37: nit: these values should be a constant enum

PR #654, opened 2026-05-02
feat: register UniswapX V3 chains (Tempo + 9 multichain)

codyborn:
## Summary

- Multichain registration in `lib/util/chain.ts`: adds BNB, Monad, XLayer, Worldchain, Soneium, Celo, Avalanche, Blast, Zora to the `ChainId` enum, plus all 9 + Optimism (previously defined but unregistered) to `SUPPORTED_CHAINS`. Tempo registered.
- `BLOCK_TIME_MS_BY_CHAIN` and `OLDEST_BLOCK_BY_CHAIN` populated per chain.
- `AVERAGE_BLOCK_TIME(chainId)` extended to all rollout chains; values cross-referenced against live RPC measurements:
  - BNB → 0.5s (post-Maxwell hardfork dropped block time mid-2025)
  - Monad → 0.5s
  - XLayer → 1s
  - Celo → 1s (chain migrated from 5s blocks late 2024)
  - Polygon → 2s (was deliberately inflated to 12s; tightened for faster fill detection)
- **`MIN_RETRY_WAIT_SECONDS = 1`** floor on `calculateDutchRetryWaitSeconds`, applied chain-agnostically. Step Functions Wait state granularity is whole seconds, so any sub-second block time would round to zero and hot-loop. Chains with ≥1s `AVERAGE_BLOCK_TIME` are unaffected by the floor.
- `lib/handlers/constants.ts`: sentinel-zero `PRIORITY_ORDER_TARGET_BLOCK_BUFFER` and `HYBRID_ORDER_TARGET_BLOCK_BUFFER` entries for chains that register `Dutch_V3` only. Unreachable in practice — `OffChainUniswapXOrderValidator.validateReactorAddress` rejects those order types when not in the SDK's reactor mapping.
- `getRpcUrl(chainId)` helper in `lib/Config.ts`: reads `RPC_<chainId>` if set (override), else `${RPC_PREFIX_URL}/${chainId}`. Throws if neither. Existing deployments keep working via the per-chain override; new chains can be enabled by setting a single `RPC_PREFIX_URL`. `CONFIG.rpcUrls` is now lazy-resolved (per-chain on first use) so imports don't require env to be fully configured at module load. `gs-reaper` and `check-order-status/util.ts:getProvider` migrated off direct `process.env` reads.
- `bin/app.ts` wires `RPC_PREFIX_URL` through both the deployed-Lambda env (sourced from `jsonRpcProvidersSecret`) and the local-dev `envVars` map (sourced from `process.env`).
- `.env.example` rewritten around `RPC_PREFIX_URL` with `RPC_<chainId>` shown as the override pattern.

## Notes for reviewer

- AWS Secrets Manager provisioning for `RPC_PREFIX_URL` is tracked separately in [ECO-374](https://linear.app/uniswap/issue/ECO-374).
- `HybridOrder` tests are blocked on `@uniswap/uniswapx-sdk@3.0.3`'s cosignerData shape change; SDK bump is held until adapter work lands ([ECO-382](https://linear.app/uniswap/issue/ECO-382)). This PR keeps the existing `3.0.0-beta.14` pin.

## Context

Part of Linear [ECO-365](https://linear.app/uniswap/issue/ECO-365). Companion PRs in x-contracts, uniswapx-sdk, x-parameterization-api, trading-api.

## Test plan

- [x] `yarn lint`: clean
- [x] `yarn build`: clean
- [x] `yarn test`: 506/506 passed
- [ ] Reviewer: per-chain `BLOCK_TIME_MS_BY_CHAIN` / `AVERAGE_BLOCK_TIME` values match measured chain reality
- [ ] Reviewer: `MIN_RETRY_WAIT_SECONDS = 1` is safe wrt. Step Functions execution-history budget on sub-second-block chains


codyborn:
## Code Review — PR #654 (x-service Tempo registration)

**Diff**: `+133/-6` across 6 files; mergeable.

### Overview
Registers Tempo (chainId `4217`) as a supported chain. The non-trivial work is `calculateDutchRetryWaitSeconds` — Tempo's 0.5s blocks would round to 0 in Step Functions Wait granularity, so a Tempo-scoped `MIN_RETRY_WAIT_SECONDS_TEMPO = 2` floor is added. Sentinel zeros are added to `PRIORITY_ORDER_TARGET_BLOCK_BUFFER` and `HYBRID_ORDER_TARGET_BLOCK_BUFFER` because those maps are typed `Record<ChainId, number>`, with comments citing the upstream guard that makes the values unreachable.

### Strengths
- **Floor scoped to Tempo only** (revised from an earlier global floor) — explicit `if (chainId === ChainId.TEMPO)` keeps Arbitrum/Unichain at their existing 1s waits. Tests assert non-Tempo chains are unaffected, locking the contract.
- Comment block above `MIN_RETRY_WAIT_SECONDS_TEMPO` explains the reasoning (SF whole-second granularity + execution-history limit) with the math worked out (300 retries × ~7 events ≈ 2,100 << 25k).
- Sentinel `[ChainId.TEMPO]: 0` for priority/hybrid buffers carries a thorough comment naming the upstream guard (`OffChainUniswapXOrderValidator.validateReactorAddress` + missing SDK reactor entry).
- Test coverage is thorough: positive (Tempo gets the floor), negative (Mainnet/Arbitrum/Unichain don't), and registration sanity (`BLOCK_TIME_MS_BY_CHAIN[TEMPO]` defined).
- Verified that `AVERAGE_BLOCK_TIME` returning `0.5` is safe for all callers (`timestampToBlockNumber` uses `Math.floor`, `HybridOrder`/`PriorityOrder` use `blockTimeSeconds * 0.75` — both fraction-safe).

### Issues / Suggestions

**`.env.example` has stale entries:**
- The new `.env.example` file lists `RPC_10=` (Optimism), but `OPTIMISM` is in the `ChainId` enum but **not** in `SUPPORTED_CHAINS` in `lib/util/chain.ts`. Since this file is created fresh in this PR, it should match `SUPPORTED_CHAINS` exactly. Either:
  1. Remove `RPC_10` from `.env.example`, or
  2. Add `ChainId.OPTIMISM` to `SUPPORTED_CHAINS` if it should be supported.

**Sentinel value choice — defense in depth:**
- `[ChainId.TEMPO]: 0` for priority/hybrid buffers relies on the upstream `validateReactorAddress` guard. If that guard ever regresses or a future change adds Tempo to the priority reactor mapping in uniswapx-sdk, a buffer of `0` means "fill at the current block" — which would behave unpredictably rather than failing loudly. Consider one of:
  1. `Number.NaN` (any arithmetic with it produces NaN, easier to trace in logs)
  2. A sentinel like `-1` plus an assertion at the call site (`assert(buffer >= 0)`)
  3. Convert the type to `Partial<Record<ChainId, number>>` and let consumers gate on presence
- Lowest-effort: add a runtime assertion at the call site of these buffers — `if (buffer === 0) throw` would catch a regression at the right layer.

**Missing test for the upstream guard claim:**
- The comments assert "Tempo priority/hybrid orders are rejected upstream because REACTOR_ADDRESS_MAPPING has no entry." Worth adding a test in `OffChainUniswapXOrderValidator` (or wherever the validator lives) that confirms a Priority order with `chainId=4217` is rejected. Right now the claim is documentation only.

**`OLDEST_BLOCK_BY_CHAIN[TEMPO] = 17850000`:**
- This was the chain height around 2026-05-02 when the Tempo work was scoped. Since this is a lookback floor, it's safe for it to be older than current — but worth a comment noting its provenance so future readers understand the value isn't load-bearing on a specific block.

**Style nit:**
- The fall-through comment on `case ChainId.TEMPO: ...` (lines 255–260 in the diff) explains why the value is 0.5 and the SF floor concern. Great. Consider also linking to `MIN_RETRY_WAIT_SECONDS_TEMPO` so a reader who lands on `AVERAGE_BLOCK_TIME` discovers the floor without grepping.

### Test coverage
- 505/505 unit tests pass per the PR description.
- New tests: 6 in `calculateDutchRetryWaitSeconds` describe blocks + 2 chain-registration sanity checks. Solid coverage for the changes.
- Gap: no test exercises `timestampToBlockNumber(refBlock, targetTimestamp, ChainId.TEMPO)` with the fractional 0.5s block time — should produce sensible block deltas. Worth adding a unit test (e.g., 30s ahead → 60 blocks ahead).

### Performance / Security
- N/A on perf (config). 
- Security: no concerns. Floor only delays retries; the upstream guard for priority/hybrid is the right belt-and-suspenders.

### Verdict
**Approve with minor revisions.** Trim the stale `RPC_10` from `.env.example`. Consider hardening the priority/hybrid sentinel (NaN or a runtime assertion). The other items are polish.

🤖 Review performed via `/review`

lib/handlers/check-order-status/util.ts
```diff
@@ -252,10 +253,39 @@ export const AVERAGE_BLOCK_TIME = (chainId: ChainId): number => {
       return 1
     case ChainId.BASE:
       return 2
+    case ChainId.OPTIMISM:
+      return 2
+    case ChainId.BNB:
+      // Post-Maxwell hardfork; calculateDutchRetryWaitSeconds floors retry
+      // waits at MIN_RETRY_WAIT_SECONDS so Step Functions Wait doesn't round
+      // to zero.
+      return 0.5
```
alanhwu: 0.45

lib/handlers/check-order-status/util.ts
```diff
@@ -252,10 +253,39 @@ export const AVERAGE_BLOCK_TIME = (chainId: ChainId): number => {
...
+    case ChainId.MONAD:
+      // Sub-second; calculateDutchRetryWaitSeconds floors retry waits at
+      // MIN_RETRY_WAIT_SECONDS so Step Functions Wait doesn't round to zero.
+      return 0.5
```
alanhwu: 0.4

lib/handlers/check-order-status/util.ts
```diff
@@ -252,10 +253,39 @@ export const AVERAGE_BLOCK_TIME = (chainId: ChainId): number => {
...
+    case ChainId.XLAYER:
+      return 1
+    case ChainId.WORLDCHAIN:
+      return 2
+    case ChainId.SONEIUM:
+      return 2
+    case ChainId.AVALANCHE:
+      return 2
```
alanhwu: 1

lib/handlers/check-order-status/util.ts
```diff
@@ -252,10 +253,39 @@ export const AVERAGE_BLOCK_TIME = (chainId: ChainId): number => {
...
+      // Reported in seconds as a fractional value for accurate block-number
+      // arithmetic (see timestampToBlockNumber); calculateDutchRetryWaitSeconds
+      // floors retry waits at MIN_RETRY_WAIT_SECONDS so Step Functions Wait
+      // state doesn't round to zero.
+      return 0.5
     case ChainId.POLYGON:
-      // Keep this at the default 12 for now since we would have to do more retries
-      // if it was at 2 seconds
-      return 12
+      return 2
```
alanhwu: 1.75

lib/handlers/check-order-status/util.ts
alanhwu: Not part of the edits but this should be 0.25 🫢

alanhwu: Just a broader comment on this for the future, I feel like we should move to SDK since we also do this mapping in GPA and it'd be good to have only 1 source of truth and place to update numbers

lib/util/chain.ts
```diff
@@ -7,16 +7,38 @@ export enum ChainId {
   POLYGON = 137,
   SEPOLIA = 11155111,
   UNICHAIN_SEPOLIA = 1301,
+  TEMPO = 4217,
```
alanhwu: Should we be grabbing this enum from SDK instead?

lib/util/chain.ts
```diff
@@ -7,16 +7,38 @@ export enum ChainId {
...
+  BLAST = 81457,
+  ZORA = 7777777,
 }
 
-// If you update SUPPORTED_CHAINS, ensure you add a corresponding RPC_${chainId} environment variable.
-// lib/config.py will require it to be defined.
+// Each chain in SUPPORTED_CHAINS needs an RPC URL resolvable by
+// getRpcUrl(chainId) (Config.ts) — either RPC_PREFIX_URL covering the
+// chain or a per-chain RPC_<chainId> override.
 export const SUPPORTED_CHAINS = [
```
alanhwu, line 8: Calling out that the OrderValidator proactively instantiates for everything we have in this list, so we'll need to make sure to have all the RPC URLs for these so it doesn't throw

lib/handlers/check-order-status/util.ts
```diff
@@ -15,6 +15,7 @@ import { BigNumber, ethers } from 'ethers'
 import { ORDER_STATUS, SettledAmount } from '../../entities'
 import { ChainId } from '../../util/chain'
 import { NATIVE_ADDRESS, RPC_HEADERS } from '../../util/constants'
+import { getRpcUrl } from '../../Config'
```
alanhwu, line 19: Need to do the new RPC fetching for the `check-order-status` injector too, it still follows the `RPC_` pattern

bin/app.ts
```diff
@@ -170,6 +170,9 @@ export class APIPipeline extends Stack {
     })
 
     const jsonRpcUrls: { [chain: string]: string } = {}
+    // Shared prefix the Lambda's getRpcUrl appends the chainId to. Per-chain
+    // RPC_<chainId> entries below override it where set.
+    jsonRpcUrls['RPC_PREFIX_URL'] = jsonRpcProvidersSecret.secretValueFromJson('RPC_PREFIX_URL').toString()
     Object.values(SUPPORTED_CHAINS).forEach((chainId) => {
       const key = `RPC_${chainId}`
```
alanhwu: May need to remove/modify this part too since it might be searching for the `RPC_` pattern during deployment before the runtime override logic

alanhwu (approved):

lib/util/chain.ts
codyborn: Good callout. We should have all these in Unimind.

lib/util/chain.ts
codyborn: yes, good catch

lib/handlers/check-order-status/util.ts
codyborn: Done

.env.example
```diff
@@ -0,0 +1,20 @@
+# RPC URL resolution: getRpcUrl(chainId) returns RPC_<chainId> if set,
+# otherwise `${RPC_PREFIX_URL}/${chainId}`. Set RPC_PREFIX_URL for the
+# common case; set per-chain RPC_<chainId> only when a chain needs a
+# different provider.
+RPC_PREFIX_URL=
+# RPC_4217=https://rpc.tempo.xyz   # example override
```
alanhwu: update comments here after the rpc loading change

alanhwu (approved):

PR #674, opened 2026-06-04
feat: send x-internal-service-secret header on all RPC requests

codyborn:
## Summary

Adds an `x-internal-service-secret` authentication header to **all** outbound RPC requests, sourced from a new `RPC_HEADER_SECRET` env var.

> **Stacked on #673** (the per-chain RPC revert). This builds on the restored `RPC_PREFIX_URL` codebase, so its base is the revert branch. Retarget to `main` once #673 merges.

## How it works

All RPC providers are constructed with the shared `RPC_HEADERS` object from `lib/util/constants.ts` (check-order-status injector/util, gs-reaper, `EventWatcherMap`, shared providers). Adding the header there wires it into every RPC request through a single change:

```ts
export const RPC_HEADERS: { [key: string]: string } = {
  'x-uni-service-id': 'x_order_service',
  ...(process.env.RPC_HEADER_SECRET ? { 'x-internal-service-secret': process.env.RPC_HEADER_SECRET } : {}),
}
```

The header is **omitted** when `RPC_HEADER_SECRET` is unset (local dev / unit tests), so behavior is unchanged where the secret isn't provided.

## CDK wiring (`bin/app.ts`)

`RPC_HEADER_SECRET` is read from the existing `gouda-service-rpc-urls-2` secret (new JSON key) and passed as a Lambda env var to:
- **beta** and **prod** stages
- the **integ-test** CodeBuild (so e2e mainnet RPC calls carry the header)
- the **local dev** stack (from `process.env`)

It is deliberately kept **out of** the `jsonRpcUrls` `CfnOutput` so the secret value is not exposed in CloudFormation outputs.

## Docs & tests

- Documented in `.env.example`, `CLAUDE.md`, `README.md`.
- New `test/unit/util/rpc-headers.test.ts` verifies the header is present when the secret is set and absent when unset.

## Required before deploy

⚠️ Add the `RPC_HEADER_SECRET` JSON key to the `gouda-service-rpc-urls-2` secret (and ensure it's present for the integ-test build) before deploying, or the `secretValueFromJson` lookups will fail.

## Testing

- `yarn build` ✅
- `yarn test` — 511 unit tests pass (3 new) ✅

### DIFF
--- a/test/unit/util/rpc-headers.test.ts
+++ b/test/unit/util/rpc-headers.test.ts
@@ -1,0 +1,33 @@
+// RPC_HEADERS reads RPC_HEADER_SECRET at module load, so each case sets the env
+// and re-imports the module in isolation.
+describe('RPC_HEADERS', () => {
+  const savedEnv = { ...process.env }
+
+  afterEach(() => {
+    process.env = { ...savedEnv }
+    jest.resetModules()
+  })
+
+  const loadHeaders = (): { [key: string]: string } => {
+    let headers: { [key: string]: string } = {}
+    jest.isolateModules(() => {
+      headers = require('../../../lib/util/constants').RPC_HEADERS
+    })
+    return headers
+  }
+
+  it('always sets the service id header', () => {
+    delete process.env.RPC_HEADER_SECRET
+    expect(loadHeaders()['x-uni-service-id']).toEqual('x_order_service')
+  })
+
+  it('adds the x-internal-service-secret header when RPC_HEADER_SECRET is set', () => {
+    process.env.RPC_HEADER_SECRET = 'super-secret-value'
+    expect(loadHeaders()['x-internal-service-secret']).toEqual('super-secret-value')
+  })
+
+  it('omits the x-internal-service-secret header when RPC_HEADER_SECRET is unset', () => {
+    delete process.env.RPC_HEADER_SECRET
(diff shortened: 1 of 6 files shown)


test/unit/util/rpc-headers.test.ts
```diff
@@ -0,0 +1,33 @@
...
+
+  afterEach(() => {
+    process.env = { ...savedEnv }
+    jest.resetModules()
+  })
+
+  const loadHeaders = (): { [key: string]: string } => {
+    let headers: { [key: string]: string } = {}
+    jest.isolateModules(() => {
+      headers = require('../../../lib/util/constants').RPC_HEADERS
```
ericneil-sanc: Maybe add this? 
`// eslint-disable-next-line @typescript-eslint/no-var-requires`

or change to
`await import('../../../lib/util/constants')`

or something and deal with all the resulting async requirements

ericneil-sanc (approved):

test/unit/util/rpc-headers.test.ts
codyborn: Good catch — `yarn lint` was indeed failing on this (`@typescript-eslint/no-var-requires`). Switched to `await import('../../../lib/util/constants')` with `jest.resetModules()` and made the test cases async. Lint is clean and all three cases still pass (the module re-evaluates per case, so the `RPC_HEADER_SECRET` branch is still exercised). Fixed in ff36bc8.

PR #678, opened 2026-06-08
feat, test: chain-based fill lookups to avoid false cancel tags

alanhwu:
Fixes false cancelled classification for filled Dutch V3 orders by widening terminal fill lookup to decayStartBlock when needed, and adds chain-based overlap constants (mainnet=20, base=20, arbitrum=50) with test coverage.

### DIFF
--- a/lib/handlers/check-order-status/service.ts
+++ b/lib/handlers/check-order-status/service.ts
@@ -21,13 +21,16 @@ import { ChainId } from '../../util/chain'
 import { metrics } from '../../util/metrics'
 import { SfnStateInputOutput } from '../base'
 import { FillEventLogger } from './fill-event-logger'
-import { getSettledAmounts, IS_TERMINAL_STATE, timestampToBlockNumber } from './util'
+import {
+  FILL_CHECK_OVERLAP_BLOCKS_ON,
+  getSettledAmounts,
+  IS_TERMINAL_STATE,
+  timestampToBlockNumber,
+} from './util'
 import { parseOrder } from '../OrderParser'
 import { PRIORITY_ORDER_TARGET_BLOCK_BUFFER } from '../constants'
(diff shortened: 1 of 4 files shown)


lib/handlers/check-order-status/service.ts
codyborn, line 134: Something like:
if (DutchV3 || Priority) {
 fromBlock = Math.max(0, Math.min(fillSearchFromBlock, order.cosignerData.decayStartBlock))
}
else {
  fromBlock = !startingBlockNumber ? curBlockNumber - this.fillEventBlockLookback(chainId) : startingBlockNumber
}

lib/handlers/check-order-status/service.ts
```diff
@@ -145,10 +148,21 @@ export class CheckOrderStatusService {
...
+    const overlapBlocks = FILL_CHECK_OVERLAP_BLOCKS_ON(chainId as ChainId)
+    let fillSearchFromBlock = Math.max(0, fromBlock - overlapBlocks)
+    if (
+      getFillLogAttempts > 0 &&
+      (validation === OrderValidation.NonceUsed || validation === OrderValidation.Expired) &&
+      order.type === OrderType.Dutch_V3
+    ) {
+      // Minimal reconciliation: avoid false terminal states for V3 by looking
+      // back to the auction start block before concluding cancelled/expired.
+      fillSearchFromBlock = Math.max(0, Math.min(fillSearchFromBlock, order.cosignerData.decayStartBlock))
```
codyborn, line 160: Curious, why don't we always set `fromBlock` to this for v3? It seems like `fillEventBlockLookback()` and `FILL_CHECK_OVERLAP_BLOCKS_ON` perform similar jobs.

alanhwu:
Fixed in #687

PR #683, opened 2026-06-14
fix: stop misclassifying filled orders as CANCELLED/EXPIRED/ERROR (reaper + SFN monitor)

codyborn:
## Summary

Orders that actually **filled** on-chain are being written to the DB as **CANCELLED** (and, in edge cases, EXPIRED/ERROR). Found while auditing BNB Dutch V3: of the orders the `/v2/orders` API reports as `cancelled`, multiple have a successful on-chain `Fill` event for their exact `orderHash` (e.g. `0x9149546f…01f32c` → fill tx `0x35a75925…`; `0x1a5da7b1…f26ac9` → fill tx `0xd4b000c7…`; the `Fill` event's `swapper` matches the order's swapper in both).

Both components that resolve order status share one flawed inference: **a used nonce ⇒ CANCELLED** (and an expired deadline ⇒ EXPIRED). But a fill *also* consumes the nonce — so whenever the fill event isn't located, a filled order is mislabeled. This PR fixes both the **GS reaper** and the **step-function monitor** (`check-order-status`).

---

## Part 1 — GS reaper (`lib/crons/gs-reaper/gs-reaper.ts`)

**Flaw:** PROCESS_BLOCKS scans block ranges for fills; CHECK_CANCELLED then marks any remaining order with `NonceUsed` as CANCELLED. When `getFillEvents` fails all `REAPER_MAX_ATTEMPTS` for a range (e.g. the RPC rejects the `BLOCK_RANGE = 10000` `getLogs` window — common on BNB), `processBlockRange` gives up and the still-OPEN order falls through to CHECK_CANCELLED, where its used nonce is misread as a cancel. A transient/structural fill-visibility gap becomes a permanent wrong terminal status.

**Fix:** track whether fill-event scanning completed for every range this run (`fillScanComplete`; `processBlockRange` now returns `fillScanFailed`). When it didn't, **defer** the `NonceUsed → CANCELLED` decision and leave the order unresolved so a later run with full fill visibility resolves it.

## Part 2 — Step-function monitor (`lib/handlers/check-order-status/service.ts`)

The SFN has its own copy of the inference (`getUnfilledStatusFromValidation`) and an even more fragile fill search:

- **Frozen, tiny window:** the fill search is anchored to the first poll's block minus `fillEventBlockLookback` (**10 blocks** default/BNB, 100 Polygon) and the lower bound is frozen for the order's life. A fill before that floor — e.g. an open/exclusive fill at or just before `decayStartBlock` (we observed fills 5–7 blocks *before* it) — is never found, and the used nonce becomes CANCELLED after a single grace poll.
- **Unguarded lookup:** `getFillInfo` wasn't wrapped, so an RPC `getLogs` range/rate error either bubbled (orphaning the order to the reaper) or let the unfilled path finalize a wrong status.
- **UnknownError → terminal ERROR:** an ambiguous/transient validator result was mapped straight to terminal ERROR, with no fill check and no grace.

**Fixes:**
1. `getFillSearchFromBlock`: anchor the search lower bound to the order's **exact** decay/auction start (Dutch V3, Hybrid, Priority) so early fills are always in range, regardless of when polling first ran; keep the rolling window for timestamp-based types (Dutch, Dutch V2); never shrink coverage below the rolling window.
2. Wrap the fill lookup; on failure keep the order **OPEN** instead of concluding CANCELLED/EXPIRED on incomplete fill visibility.
3. Route `UnknownError` to **OPEN** (keep polling) instead of terminal ERROR; genuine `InvalidSignature`/`InvalidOrderFields` still map to ERROR.

---

## Testing

- `test/unit/handlers/gs-reaper/gs-reaper.test.ts` — 18/18 (incl. 2 new: defer-on-incomplete-scan, and PROCESS_BLOCKS flags `fillScanComplete=false` when a range exhausts retries).
- `test/integ/handlers/check-order-status-service.test.ts` — 18/18 (incl. 4 new: window anchoring for V3/Priority/Dutch + no-shrink; fill-lookup failure → open not cancelled at `getFillLogAttempts=1`).
- `test/unit/services/check-order-status/CheckOrderStatusUtils.test.ts` — UnknownError now asserts `open`.
- `tsc --noEmit` clean; eslint 0 errors.

Across both: deferring (an order stays OPEN briefly) is strictly safer than emitting a wrong terminal status.

## Follow-ups (not in this PR)

- The reaper's `BLOCK_RANGE = 10000` `getLogs` window likely exceeds some chains' RPC limits (the reason scans fail on BNB) — worth making per-chain / chunked so fills are actually *found*, not just deferred.
- Backfill/repair orders already mislabeled CANCELLED that have on-chain fills.


lib/handlers/check-order-status/service.ts
```diff
@@ -242,6 +272,42 @@ export class CheckOrderStatusService {
...
+   * anchor to it so fills that land at or just before it are always in range,
+   * regardless of when polling first ran. For timestamp-based order types
+   * (Dutch, Dutch V2) we keep the rolling lookback window. Always at least
+   * FILL_CHECK_OVERLAP_BLOCK below the rolling window so coverage never shrinks.
+   */
+  private getFillSearchFromBlock(
+    order: UniswapXOrderEntity,
+    chainId: number,
+    rollingFromBlock: number
+  ): number {
```
ericneil-sanc, line 314: Can we make this handle timestamp-based orders, because dont we still run into an issue if the order is filled before the polling window for those?

lib/handlers/check-order-status/service.ts
```diff
@@ -366,8 +432,12 @@ export class CheckOrderStatusUtils {
         }
       case OrderValidation.InvalidSignature:
       case OrderValidation.InvalidOrderFields:
-      case OrderValidation.UnknownError:
         return { orderStatus: ORDER_STATUS.ERROR }
+      case OrderValidation.UnknownError:
+        // Ambiguous/transient validator result (e.g. an unrecognized revert or a
+        // flaky RPC). Don't finalize as terminal ERROR -- the order may be valid
+        // or already filled. Keep polling so a later run can resolve it.
+        return { orderStatus: ORDER_STATUS.OPEN }
```
ericneil-sanc: This could result in an infinte loop - the existing retry limit of 300 spins up kickoffOrderTrackingSfn

ericneil-sanc (approved):

PR #692, opened 2026-08-06
fix, test: stop writing client IPs to CloudWatch logs

codyborn:
## What

The API handler logged the raw `APIGatewayProxyEvent` in two places. That event carries the client IP in several fields — `requestContext.identity.sourceIp` and the forwarding headers (`X-Forwarded-For`, `X-Real-IP`, `CF-Connecting-IP`) — so every hit of the error path wrote an IP into CloudWatch, where prod log groups retain for **three months** (`logRetentionDays`, `lib/util/stage.ts`).

The error-level call is the notable one: it runs in prod regardless of `NODE_ENV`, unlike the debug call that only fires under `NODE_ENV=debug` (which nothing in `bin/` sets).

```
lib/handlers/base/api-handler.ts:118  log.debug({ event, context }, 'Request started.')
lib/handlers/base/api-handler.ts:152  log.error({ err, event }, 'Unexpected error building request injected.')
```

## How

Added `redactEvent()` and applied it at both sites. It's an **allowlist rather than a denylist**, so a new IP-bearing header can't leak in by default.

- **Kept:** `resource`, `path`, `httpMethod`, `pathParameters`, `queryStringParameters`, `body`, `isBase64Encoded`, and exactly six `requestContext` fields — `requestId`, `path`, `stage`, `resourcePath`, `httpMethod`, `requestTimeEpoch`.
- **Dropped:** `headers`, `multiValueHeaders`, and every other `requestContext` field. That includes `identity` (which carries `sourceIp`, `userAgent`, `caller`, `user`) and also `authorizer`, `apiId`, `extendedRequestId` and the domain fields. Dropping `authorizer` matters independently of IPs, since it can carry claims.

This brings the Lambda logs in line with the API Gateway access logs, which already set `ip: false` in the access log format (`bin/stacks/api-stack.ts:88`).

## Context

Came out of a review of the WAF IP rate limiting in `bin/stacks/api-stack.ts`. IP addresses are personal data under GDPR (Recital 30; CJEU *Breyer*, C-582/14), and the WAF rate limiting itself is a standard legitimate-interest use for abuse prevention. But months-long IP retention in application logs is harder to justify as necessary, and it was inadvertent rather than deliberate.

Two related surfaces were left alone deliberately, as they're judgment calls rather than clear bugs:

- `sampledRequestsEnabled: true` (7×) — retains sampled requests including Source IP for 3 hours. Genuinely useful when debugging a misfiring WAF rule.
- **Request bodies are still logged**, which for order posts means a signed order. Note this error path fires inside `getRequestInjected`, i.e. *before* the order is persisted, so the log entry can be the only copy — the "it's in DynamoDB anyway" argument doesn't actually cover this call site. Keeping the body still looks right for debuggability, since the signature only authorizes the order the client already intended to submit, but it's a deliberate call rather than a free one.

## Testing

New `test/unit/handlers/base/redact-event.test.ts` asserts the exact output shape and serializes the result to confirm no IP (or `x-api-key`) survives — so it fails loudly if someone re-adds a field later.

Uses `toStrictEqual` rather than `toEqual`, since `toEqual` ignores keys whose value is `undefined` and would miss a re-added field absent from the fixture. Verified both ways: adding `multiValueQueryStringParameters` to `redactEvent` fails under `toStrictEqual` and passes under `toEqual`.

- `npx jest test/unit` — 43 suites / 508 tests pass
- `npx tsc --noEmit` — clean
- `npx eslint` on both files — clean

### DIFF
--- a/lib/handlers/base/api-handler.ts
+++ b/lib/handlers/base/api-handler.ts
@@ -23,6 +23,32 @@ export const INTERNAL_ERROR = (id?: string) => {
 
 export type APIGatewayProxyHandler = (event: APIGatewayProxyEvent, context: Context) => Promise<APIGatewayProxyResult>
 
+/*
+ * The raw APIGatewayProxyEvent carries the client IP in several places: requestContext.identity.sourceIp
+ * and any of the forwarding headers (X-Forwarded-For, X-Real-IP, CF-Connecting-IP, ...). IP addresses are
+ * personal data, and our log groups retain for up to three months, so never log the raw event.
+ * This is an allowlist rather than a denylist so a new IP-bearing header can't leak in by default.
+ */
+export function redactEvent(event: APIGatewayProxyEvent) {
+  return {
+    resource: event.resource,
+    path: event.path,
+    httpMethod: event.httpMethod,
+    pathParameters: event.pathParameters,
+    queryStringParameters: event.queryStringParameters,
+    body: event.body,
+    isBase64Encoded: event.isBase64Encoded,
+    requestContext: {
+      requestId: event.requestContext?.requestId,
+      path: event.requestContext?.path,
+      stage: event.requestContext?.stage,
+      resourcePath: event.requestContext?.resourcePath,
+      httpMethod: event.requestContext?.httpMethod,
+      requestTimeEpoch: event.requestContext?.requestTimeEpoch,
+    },
+  }
+}
+
 export type ApiRInj = {
   log: Logger
   requestId: string
@@ -115,7 +141,7 @@ export abstract class APIGLambdaHandler<CInj, RInj extends ApiRInj, ReqBody, Req
           requestId: context.awsRequestId,
         })
 
-        log.debug({ event, context }, 'Request started.')
+        log.debug({ event: redactEvent(event), context }, 'Request started.')
 
         let requestBody: ReqBody
         let requestQueryParams: ReqQueryParams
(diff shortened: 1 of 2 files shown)


gretzke (approved):
I checked out the branch and ran the new test file plus npx tsc --noEmit, both clean. I also grepped the repo to confirm these two call sites are the only places the raw event gets logged. The allowlist approach looks right. Three small things.

1. Consider toStrictEqual instead of toEqual in the shape test. toEqual ignores keys whose value is undefined, so the test only fails for re-added fields that also exist in the fixture. If someone adds multiValueQueryStringParameters to the allowlist, for example, it would still pass.

2. The "persisted in DynamoDB by design" rationale doesn't fully cover this call site. The error fires in getRequestInjected before the order is persisted, so the log entry can end up being the only copy of the signed order. Keeping the body for debuggability still seems like the right call, since the signature only allows submitting the order the client already wanted submitted. Just noting the description overstates the justification a bit.

3. The description says the non-identity requestContext fields are kept, but the helper keeps only six of them and drops authorizer, apiId, extendedRequestId and the domain fields. That is a good thing since authorizer can carry claims, so if anything the description undersells the allowlist.

codyborn:
Thanks — all three taken.

1. **`toStrictEqual`** — done in 121ef86, and you're right about the failure mode. I verified it both ways: adding `multiValueQueryStringParameters` to `redactEvent` (absent from the fixture, so `undefined`) fails the test under `toStrictEqual` and passes under `toEqual`. Added a comment above the assertion so the choice doesn't get reverted as noise.

2. **The DynamoDB rationale** — you're right that it doesn't cover this call site, since the throw happens inside `getRequestInjected` before the order is persisted. Corrected in the description: the body is still kept, but now stated as a deliberate call justified by the signature only authorizing the order the client already intended to submit, rather than by "it's in Dynamo anyway."

3. **`requestContext` wording** — also correct, the helper keeps six named fields and drops the rest including `authorizer`. Description now lists the kept fields explicitly and calls out `authorizer` as worth dropping on its own merits.

No code change for 2 or 3 — both were description accuracy.

ericneil-sanc (approved):

PR #696, opened 2026-08-24
fix: cut Get Orders read amplification on hot GSI partitions

codyborn:
## Problem

`GET /dutch-auction/orders` is throwing **"Throughput exceeds the current capacity of your table or index"** in prod.

The Orders table is `PAY_PER_REQUEST` (`bin/config.ts:20-22`), so there is no table-level read ceiling to exceed. The limit actually being hit is the **~3000 RCU/s DynamoDB caps a single partition at**, and filler polling concentrates on exactly one partition key:

| Query | Index | Partition key | Distinct keys |
|---|---|---|---|
| `?chainId=1&orderStatus=open` | `chainId_orderStatus-createdAt-all` | `"1_open"` | 1 |
| `?orderStatus=open` | `orderStatus-createdAt-all` | `"open"` | 1 |

Every GSI is `ProjectionType.ALL`, so each index item carries the full order including `encodedOrder` and `signature`. A 50-item page costs roughly 12-25 RCU, which alone puts the hot key at ~120-240 queries/s.

### 250ms cache over the list-query path

New `QueryCache`, wired into `queryOrderEntity` — the single funnel every list query passes through.

- **Key** covers table name, index, partition key, effective limit, cursor, sort key/comparison, direction, and type filters. A hit is byte-identical to what the caller would have read, just up to 250ms old.
- **Bounded** (1000 entries, TTL-swept). Keys embed caller-supplied values (swapper, pair, cursor), so an unbounded map would leak in a long-lived execution environment.
- **Copy on return**, so nothing downstream can mutate a cached entry.
- **Per-endpoint instances, constructed at the composition root.** `GET /orders` and `GET /limit-orders` each build their own cache in a handler-local `query-cache.ts` (the limit-orders lambda polls the same kind of hot single-status partition on the LimitOrders table, so it gets the same protection). Metric names come from the instance — `GetOrdersQueryCacheHit/Miss` and `GetLimitOrdersQueryCacheHit/Miss` — so per-endpoint hit rates stay distinguishable on dashboards, and the generic repository carries no endpoint-specific naming.
- **Kill-switch-aware metrics.** With `GET_ORDERS_CACHE_TTL_MS=0`, the repository skips cache bookkeeping entirely (no key building, no metrics) instead of emitting a 100% miss rate that would be indistinguishable from a broken cache mid-rollback.

**The cache is opt-in per repository, not a default.** The unimind cron (`lib/crons/unimind-algorithm.ts:145,267`) writes orders and immediately re-reads them expecting fresh data — a global cache broke six of its integ tests. Only the two read-only GET lambdas pass a cache in; post-order, check-order-status, the crons, and the reaper build repositories without one. There is a test pinning that.

### Multi-status responses dedupe by orderHash

`getOrdersForMultipleStatuses` fans out one sub-query per status and merges. The sub-queries were already independent reads, and with the cache they are also independent cache entries — one can be a ≤250ms-stale hit while another reads fresh, returning the same order under two statuses. The merge now dedupes by `orderHash`, keeping the non-`open` copy (status transitions only flow away from `open`, so that copy is the newer read).

### Retry loop consolidated (and left at 10)

An earlier revision of this PR cut `MAX_QUERY_RETRY` from 10 to 3 to bound read amplification from the post-query `orderType` filter. With the cache absorbing repeats, that cut isn't needed — and since cached pages get re-served many times, full pages beat cheap short ones. The budget stays at 10, and the five per-type copies of the query-retry loop are extracted into a single `fetchOrderPages` helper.

## Expected effect

Lambda serves one request at a time per execution environment, so concurrency ≈ rate × duration and queries/sec ≈ concurrency / TTL. The reduction on a single hot key works out to roughly **TTL ÷ handler latency** — about **12x** at 250ms and ~20ms latency, independent of request rate.

**This is not a hard global ceiling.** It's a per-container cache: it scales pressure down by a large constant but does not cap it. A true cap needs a shared cache or reserved concurrency on the Lambda. The hit/miss metrics will show which is actually needed.

## Rollback

TTL is `GET_ORDERS_CACHE_TTL_MS`, default 250 (now documented in CLAUDE.md). Setting it to `0` on the Lambda disables the cache — and silences its metrics — with no code change; deliberately not hardcoded into the CDK env so a console override survives.

## Testing

- `tsc --noEmit` clean
- Unit: **585 passed / 52 suites**, including 17 new tests (8 for the cache primitive, 9 for repository behavior: TTL expiry, key isolation across partition keys / limits / tables, copy-on-return, the uncached default path, per-instance metric naming, kill-switch metric silence, and multi-status dedupe)
- Integ: matches `main` exactly. Two repository suites fail in full-suite runs both before and after this change — pre-existing cross-suite contention on dynamodb-local; both pass in isolation (63 passed). Verified by stashing and re-running on `main`.
- `eslint` 0 errors; `prettier --write` on touched files (which incidentally reformatted a few pre-existing lines)

## Not addressed here

Follow-ups from the same investigation, in rough priority order:

- Enable Contributor Insights on the Orders table and the `chainId_orderStatus` GSI to confirm the hot key from data rather than inference
- Reserved concurrency on the Get Orders Lambda as a hard ceiling on DynamoDB QPS
- Narrow-projection GSI for the polling path so `encodedOrder` isn't read on every poll
- Shard the hot partition key (`1_open_<0..N>`) to raise the per-key ceiling itself
- `PROD_INDEX_CAPACITY` (`bin/config.ts:4-17`) is dead config — never passed to `APIStage`
- `THROTTLE_PER_FIVE_MINS=3000` overrides all five WAF rules, loosening `ip-post-order` 10x from its 300 default
- The WAF `allow-api-key` rule exempts keyed callers from all rate limiting on this endpoint


codyborn:
Addressed the review in 231ac4c. Four code fixes plus one correction to this PR's description.

**1. Cache entries were dated from before the round trip — real bug, fixed.**

`Date.now()` was captured once and reused for both the lookup and the write, so `expiresAt = queryStart + 250ms`. Any query taking ≥250ms wrote an entry that was already expired. The failure mode is self-reinforcing and exactly backwards: once the `1_open` partition starts throttling, aws-sdk v2 retries with backoff, latency climbs past the TTL, and the cache silently stops absorbing load precisely when it is needed — while still paying insert and eviction cost. Now stamped from completion. There's a regression test (`dates an entry from when the query finished, not when it started`) that I verified fails against the previous behaviour.

**2. The documented kill switch did not survive a deploy — my PR description was wrong.**

I claimed `GET_ORDERS_CACHE_TTL_MS` was deliberately left out of CDK "so a console override survives." It doesn't: `bin/app.ts` builds `envVars` as an explicit allowlist, `getOrdersEnv` spreads it into `NodejsFunction.environment`, and CDK writes the *complete* `Environment.Variables` map on every deploy, dropping anything it doesn't know about.

It's now plumbed through the beta and prod stage envVars at `'250'`, and through the local dev passthrough. **Corrected rollback:** set it to `'0'` and deploy. A console edit still works as an emergency lever, but only until the next deploy resets it.

**3. Retry reduction is now measurable rather than assumed.**

`MAX_QUERY_RETRY` 10 → 3 does shrink per-poll completeness for sparse order types behind a dense partition — a real tradeoff, and clients that don't follow the cursor will see it. Rather than guess, all five loops now emit `GetOrdersRetryExhausted` (with the order type logged) when they give up with a cursor still outstanding. If that fires at any volume after deploy, the right fix is folding `type` into the index key so it stops being a post-query filter, not raising the retry budget back up.

**4. `copyQueryResult` comment overpromised.**

It said mutation "can never reach a cached entry"; the copy is shallow, so that's true of the array but not the order objects. No current caller mutates them — the `offerer`→`swapper` rewrite in `handler.ts` rebuilds rather than mutates — but the comment would have read as a guarantee to whoever adds an in-place tweak next. Reworded to state what it actually protects and that entities are read-only. Left it shallow deliberately: deep-copying 50 full orders per request would burn real CPU to defend against a hypothetical.

**5. Tests no longer depend on the ambient environment.**

The suite built on the exported singleton, whose TTL reads `GET_ORDERS_CACHE_TTL_MS` — so exporting `0` in a shell (now a more likely thing to do) would have failed five tests with a confusing "expected 1 call, got 2". They construct their own `QueryCache(250)`.

**Verification:** `tsc --noEmit` clean; unit **582 passed / 52 suites**; eslint 0 errors; prettier applied. Integ still matches `main` — the repository suites fail only in full-suite runs from pre-existing dynamodb-local contention (which suite loses the race varies between runs) and pass in isolation at 63/63. The `tsconfig.cdk.json` `.env.js` error is also pre-existing; I confirmed it on a stashed baseline.

I did not change the `MAX_QUERY_RETRY` value itself — that was the requested tradeoff, and the new metric is what tells us whether it cost anything.

lib/handlers/get-orders/index.ts
```diff
@@ -13,17 +14,19 @@ import { ONE_DAY_IN_SECONDS } from '../../util/constants'
...
 import { OnChainValidatorMap } from '../OnChainValidatorMap'
 import { getMaxOpenOrders } from '../post-order/injector'
-import { DynamoQuoteMetadataRepository } from '../../repositories/quote-metadata-repository'
 
-const repo = DutchOrdersRepository.create(new DynamoDB.DocumentClient())
-const limitRepo = LimitOrdersRepository.create(new DynamoDB.DocumentClient())
+// This Lambda only reads, so its repositories share the sub-second query cache. Every
+// other entry point builds them without it -- see getOrdersQueryCache.
+const repo = DutchOrdersRepository.create(new DynamoDB.DocumentClient(), getOrdersQueryCache)
+const limitRepo = LimitOrdersRepository.create(new DynamoDB.DocumentClient(), getOrdersQueryCache)
```
gretzke, line 29: The get-limit-orders Lambda builds LimitOrdersRepository and RelayOrderRepository without the cache, so GET /limit-orders keeps the uncached read pattern on the same GSI. The same query through GET /orders?orderType=Limit is cached after this PR. If that exclusion is intentional, a note there would help. Otherwise it should probably get the same treatment.

lib/repositories/generic-orders-repository.ts
```diff
@@ -266,12 +287,37 @@ export abstract class GenericOrdersRepository<
...
+    ])
+    const queryStartedAt = Date.now()
+
+    if (this.queryCache) {
+      const cached = this.queryCache.get(cacheKey, queryStartedAt)
+      if (cached) {
+        metrics.putMetric('GetOrdersQueryCacheHit', 1, Unit.Count)
+        return this.copyQueryResult(cached as QueryResult<T>)
+      }
+      metrics.putMetric('GetOrdersQueryCacheMiss', 1, Unit.Count)
```
gretzke: With GET_ORDERS_CACHE_TTL_MS=0 the singleton is still constructed, so this branch runs and emits a miss for every query while get() always returns undefined. A dashboard then shows a 0% hit rate, which looks like a broken cache rather than a disabled one, exactly in the situation the kill switch is for. Guarding on this.queryCache?.enabled (or exporting undefined when the TTL is 0) keeps the metrics quiet when the cache is off.

gretzke:
Since the branch is getting rewritten to drop the retry change anyway, can you also remove the Claude co-author trailers from the two commits and the generated-with footer from the PR body?

alanhwu (dismissed):

alanhwu (approved):

PR #702, opened 2026-09-02
fix: collapse Get Orders cache cardinality, cap concurrency, fail fast on throttles

codyborn:
## Problem

#696 put a 250ms per-execution-environment cache in front of the Get Orders list queries and cut DynamoDB reads ~5x. Traffic then grew ~5x and the hot `1_open` GSI partition hit its per-partition limit again. Two reasons:

1. **Cache cardinality.** The key covered limit, cursor, sort and orderType filter, so every distinct combination fillers polled was its own DynamoDB read. A `cursor` changes every time an order lands, and a `createdAt=gt(...)` "since" filter changes on every poll.
2. **No ceiling.** The cache is per environment, so reads on a hot partition scale as `environments / TTL`. Lambda scales environments with demand, and when the partition throttles, latency rises, concurrency rises, more environments each miss once per TTL, and the account concurrency limit (10k) is eventually exhausted, taking post-order and status tracking down too. Get Orders alone runs ~2.5k concurrent when healthy.

## Client-visible changes (read this if you run a filler)

- **GET /orders returns one page of the newest orders** (`createdAt` descending, at most 50). `sortKey=createdAt`, `sort=gt(0)` and `desc=true` are accepted (they describe that page) and dropped. **Any other `sort`/`desc` value, or any `cursor`, is a 400** whose message names the single-page contract, so an integrator that actually relied on paging or another ordering fails loudly instead of silently getting a different result set. The response never carries a cursor. To enumerate limit orders use GET /limit-orders, not `GET /orders?orderType=Limit`.
- **GET /limit-orders is unchanged**: it keeps cursor pagination and `sortKey`/`sort`/`desc`.
- **Default ordering is newest first** on every list query (it was oldest first when no `sortKey` was given). Legacy 2023 orders with millisecond `createdAt` no longer appear in list results; they are still returned by hash.
- `orderStatus` rejects repeated values (`open,open`). `limit` must be a non-negative integer (0 = default). `pair` is bounded to 128 characters of `[0-9a-zA-Z._:-]`.
- Per-IP rate limit on `orders` paths is **1200 per 5 minutes (4 req/s)**. Requests past the Get Orders reserved concurrency surface as 5xx.

### GET /orders is a single page
The single-page schema (`GetOrdersQueryParamsJoi`) accepts `sortKey`, `sort` and `desc` only at their default values and strips them (`.strip()`), rejects every other value with a message pointing at the contract, and forbids `cursor`. The handler also ignores an injected cursor and strips any cursor from the response (`paginated: false`), and its `UniswapXOrderService` is built with `SINGLE_PAGE` so the type-filter retry loop never follows a cursor into further reads. GET /limit-orders uses `GetLimitOrdersQueryParamsJoi` and `paginated: true`.

### One cache entry per partition
Key is `(table, index, partitionKey)`. The cached path reads the newest `CACHED_QUERY_PAGE_SIZE` rows (= `MAX_ORDERS`, 50, the most the API returns) once and applies the caller's limit and orderType filter in memory, so every limit and type variant shares one read. A server-side FilterExpression would cost the same RCUs (DynamoDB bills rows evaluated), so nothing is lost.

**Only enum-validated partitions are cached**: `orderStatus`, `chainId`, `chainId_orderStatus`. `filler`, `swapper` and `pair` are caller-controlled and unbounded in cardinality, so they never enter a cache key. Instead:

- On the small statuses (`open`, `insufficient-funds`) a request naming a filler or swapper is answered from the cached page of the enum partition it sits in, filtered in memory. `chainId=1&orderStatus=open&filler=X` is served from the `1_open` page, so exclusive-filler polling costs no per-filler GSI read. The filter compares the same compound attribute the caller GSI is keyed on, so results are identical to the GSI query.
- If that base page is truncated (the partition holds more than one page), the request falls back to the caller's own GSI and emits `BaseTruncated`.
- Terminal statuses (`filled`, `expired`, ...) and caller-only queries (`swapper` history, `pair`) read their own GSI directly (`Uncacheable` metric): filtering the newest page of an unbounded partition would silently drop older history.

`orderHash` lookups never touch the cache.

**Cursors from a cached page.** The repository builds a continuation cursor when more rows exist, keyed on whichever GSI the request targets, so GET /limit-orders can follow it; following a cursor reads DynamoDB directly. When every match on a truncated page was handed back, the cursor is DynamoDB's own key so the next read does not re-scan the page. Integ tests page through DynamoDB local from a cached page, with and without a filler filter, to prove the synthetic key is accepted as `ExclusiveStartKey`.

### Newest first everywhere
List queries default to `createdAt` descending on every path. DynamoDB's default is ascending, which handed pollers the oldest page of a busy partition. The GS reaper asks for oldest-first explicitly: its persisted-cursor sweep must reach stuck (old) orders before a restart resets the cursor. The unimind cron already passed `desc: true`. Integ fixtures asserting the old order were updated.

**Legacy millisecond timestamps.** 2023 DutchLimit V1 orders stored `createdAt` in milliseconds, so numerically they sort above every real order and would head every newest-first page (the data-api's `between(...)` window and protocols-tool's `lt(1e11)` both existed to dodge this). Unless the caller supplies its own sort comparison, list reads on both paths now bound the sort key below `100000000000` as a key condition (no RCU cost; seconds stay under it until ~year 5138). Legacy rows remain reachable by hash. The bound applies to the GS reaper's sweep as well: legacy millisecond-timestamp orders are deliberately outside it. The reaper still asks for oldest-first explicitly.

### Input bounds
`limit` is validated as a non-negative integer and clamped in the repository (a negative value used to reach `Array.slice`). `orderStatus` rejects repeats: each status fans out into its own query, so `filled,filled,...` hundreds of times in one URL was a read multiplier. `pair` is bounded in length and character set; its stored value comes from free-form quote metadata, so the exact shape is not enforced.

### Observability
Per cache instance (`GetOrdersQueryCache*`, `GetLimitOrdersQueryCache*`): `Hit`, `Miss`, `Size` (Max = live distinct keys per environment), `CapacityEviction`, `Uncacheable`, `Truncated` (a cached partition exceeded one page; alarm-worthy for `open`), `BaseTruncated`. Every miss logs `table`, `index`, `partitionKey`, `rows`, `truncated`, `cacheSize`:

```
filter msg = "Query cache miss" | stats count(*) by partitionKey | sort desc
```

### Fail fast on throttles
The two GET lambdas build their DocumentClient with `maxRetries: 2` (`lib/handlers/shared/dynamo.ts`). aws-sdk v2's default of 10 exponential retries let one throttled read hold an environment for seconds, which is the feedback loop above.

### Reserved concurrency on Get Orders
`getOrdersReservedConcurrency`: **3000 prod / 100 beta**. Carved out of the account pool, so a Get Orders poll storm can no longer exhaust the 10k account limit and throttle post-order and status tracking with it. It also bounds hot-partition reads at roughly `reserved / TTL`. Get Orders runs ~2.5k concurrent when healthy, so 3000 leaves ~20% headroom while stopping the climb well short of the account limit.

### WAF
`ip-get-orders` is decoupled from `THROTTLE_PER_FIVE_MINS` (which had loosened it to 3000) and set to **1200 per 5 minutes**. With a 500ms TTL, polling any one query faster than 2 req/s returns identical data. API-key holders remain exempt via `allow-api-key`.

### Config and docs
The 500ms `GET_ORDERS_CACHE_TTL_MS` landed on main separately in #701, which is merged into this branch (its four TTL tests are included in the counts below). Swagger, README and CLAUDE.md updated; the swagger drift test checks each endpoint against its own schema.

## Expected effect
Distinct cache keys per environment drop from unbounded to roughly `tables x (statuses + chains + chains x statuses)`, a few hundred, none attacker-controllable. Reads per hot partition per environment drop to at most one per TTL regardless of how varied the polling is, and exclusive-filler polls on open orders stop reading DynamoDB at all. The reserved cap contains the blast radius of a throttle spiral. It does not remove the ceiling: at thousands of environments the per-environment cache alone cannot hold `1_open` under the partition limit, and a shared cache (DAX/Redis) or a narrow-projection GSI remains the structural fix.

## Known gaps, deliberately not addressed here
- **GET /limit-orders cursors are plain base64 and forgeable**, and any cursor (or `sort=gt(N)`) request is an uncached read on the LimitOrders `open` partition. Signed, expiring cursors with cached cursor pages were considered and deferred.
- **The WAF per-IP limits aggregate on the first `X-Forwarded-For` value** on an edge-optimized API, so they can be sidestepped by prepending a fake address. Every per-caller bound in this PR inherits that.
- Per-caller GSI reads (`swapper`/`filler`/`pair` on terminal statuses, `orderHashes`) are a billing vector on the pay-per-request table, bounded only by the WAF.

## Testing
- `tsc --noEmit` clean, eslint 0 errors
- Unit: **651 passed / 53 suites**, covering the single-page contract (default sort params stripped, non-default values and cursors rejected, cursor never returned or followed), limit-orders passthrough and its validation rules, cache key collapse, in-memory limit/type/filler/swapper filtering, truncation fallback, synthetic cursors, uncacheable indexes, limit clamping, duplicate statuses, metric names, miss log, kill-switch silence
- Integ (dynamodb-local): **185 passed / 5 suites**, including paging a real partition from a cached first page with and without a filler filter, checking the in-memory filler filter matches the GSI, and excluding a legacy ms-timestamp order from every list path while keeping it reachable by hash
- `cdk synth` of the pipeline stack: reserved concurrency 3000/100 and WAF limit 1200 present in the synthesized templates

## Rollback
`GET_ORDERS_CACHE_TTL_MS=0` disables the cache. Reserved concurrency and the WAF limit are plain CDK props in `bin/app.ts`.


bin/stacks/api-stack.ts
```diff
@@ -194,9 +200,12 @@ export class APIStack extends cdk.Stack {
...
             rateBasedStatement: {
-              // rate limit across all chains
-              // default evaluation window: 5 minutes => 6 per second
-              limit: throttlingOverride ? parseInt(throttlingOverride) : 1800,
+              // Per IP, across all chains and query shapes, over WAF's 5-minute window
+              // (1200 => 4 req/s). Tuned on its own rather than through THROTTLE_PER_FIVE_MINS:
+              // the handler serves each query from a page cached for GET_ORDERS_CACHE_TTL_MS,
+              // so polling any one query faster than that returns identical data and only
+              // spends capacity. Keyed callers are exempt via allow-api-key above.
+              limit: getOrdersThrottlePerFiveMins ?? DEFAULT_GET_ORDERS_THROTTLE_PER_FIVE_MINS,
```
ericneil-sanc, line 208: This limit is per IP across every query shape, and the uriPath CONTAINS 'orders' scope-down below also matches /limit-orders. The 2 req/s reasoning is per query. A keyless filler polling 6 chains on both endpoints at 1 Hz is 12 req/s and gets 429s here.

lib/repositories/generic-orders-repository.ts
```diff
@@ -271,77 +309,124 @@ export abstract class GenericOrdersRepository<
...
+      comparison = parseComparisonFilter(sort)
+    }
     const queryResult = await this.entity.query(partitionKey, {
       filters: filters,
       index: formattedIndex,
       execute: true,
       limit: effectiveLimit,
+      // Newest first unless a caller explicitly asks otherwise. DynamoDB's own default is
+      // ascending, which would hand a poller the oldest page of a busy partition.
+      reverse: desc,
```
ericneil-sanc, line 355: This flips the default for every uncached caller, not just the GET handlers. The reaper at lib/crons/gs-reaper/gs-reaper.ts:596 walks open orders with no sortKey, a persisted cursor and a maxOrders cap, so it now processes the newest open orders first and reaches the oldest, most likely stale, ones last. Will that break anything?

bin/stacks/api-stack.ts
codyborn: First of all, I've never heard 1 Hz used instead of 1 RPS. Always keeping it fresh @ericneil-sanc 😆 

1200 requests over 5 mins gives 4 RPS. Fillers can query all orders and then filter by chain client side so they can still get 4 RPS.

lib/repositories/generic-orders-repository.ts
codyborn: Good catch. Addressed in 9d7a10f: the reaper now passes `desc: false` explicitly (lib/crons/gs-reaper/gs-reaper.ts, getUnresolvedOrderHashes), so its sweep stays oldest-first and reaches stale orders before a restart can reset the persisted cursor. The unimind cron already passed `desc: true`, and the GET lambdas are the only other list readers, so newest-first only changes the polling endpoints.

One related note: list reads now also bound `createdAt < 1e11` by default to keep the 2023 V1 orders that stored millisecond timestamps out of newest-first pages. That bound applies to the reaper's sweep too; those legacy rows are deliberately outside it (4044847). Net for the reaper vs main: same rows, same order, same page size, minus the legacy millisecond rows. A unit test pins the reaper's query shape (uncached, ascending, bounded).

ericneil-sanc (approved):

PR #703, opened 2026-09-03
feat: dashboard section for the Get Orders cache and capacity controls

codyborn:
## What

Adds a **Get Orders Cache & Capacity** section to the bottom of the `GoudaServiceDashboard` with the metrics introduced in #702, so the cache's behaviour and the three ceilings that contain a throttle spiral are visible in one place.

| Widget | Metrics | Read it as |
|---|---|---|
| Query Cache Hit Rate | `GetOrdersQueryCacheHit/Miss`, `GetLimitOrdersQueryCacheHit/Miss` | Share of list reads served from memory, per endpoint |
| Query Cache Reads by Outcome | `Hit`, `Miss`, `Uncacheable`, `BaseTruncated` | Miss = one DynamoDB read of a cached partition. Uncacheable = bypassed the cache (cursor, sort, caller-keyed partition). BaseTruncated = filler/swapper query that fell back to its own GSI |
| Query Cache Live Keys per Environment (max) | `GetOrdersQueryCacheSize`, `GetLimitOrdersQueryCacheSize` | Distinct live keys in one execution environment. Bounded by enum partitions; a climb means the key space leaked |
| Truncated Pages & Capacity Evictions | `Truncated`, `CapacityEviction` | Truncated = a partition holds more rows than one page, so the single-page contract is hiding rows. Alarm-worthy for `open` |
| Get Orders Lambda Concurrency & Throttles | `AWS/Lambda ConcurrentExecutions` (max), `Throttles` | Reserved concurrency drawn as a horizontal line (3000 prod / 100 beta) |
| Hot GSI Consumed Read Capacity | `AWS/DynamoDB ConsumedReadCapacityUnits` on the three cacheable GSIs (`chainId_orderStatus`, `orderStatus`, `chainId`, each `-createdAt-all`), Orders + LimitOrders | Load on the partitions that throttled in INC-389 |
| Hot GSI Read Throttles | `AWS/DynamoDB ReadThrottleEvents`, same GSIs | The failure this all exists to prevent |
| WAF ip-get-orders Blocked vs Requests Served | `AWS/WAFV2 BlockedRequests` for the rule, `GetOrdersRequest` as the denominator | How often the per-IP limit bites. A block rule never emits AllowedRequests, so the service's own request count stands in |

`DashboardStack` takes an optional `getOrdersReservedConcurrency`, passed from `api-stack.ts`, for the annotation line.

## Testing

- New `test/unit/stacks/dashboard-stack.test.ts` pins the metric names, the reserved-concurrency annotation, the Maximum statistic on the cache-size widget, and the GSI and WAF dimensions. It also requires every widget to fit the 24-column grid with **zero overlaps** and the cache charts to be the only widgets under their section header.
- The legacy layout had one collision: "Orders Filled" spanned the row shared by the two "... by Chain" widgets at y=32, so CloudWatch reflowed everything beneath it. Those two move to their own row and everything below shifts down six rows, so the new section (starting at y=80) is a clean block with nothing from earlier sections reaching into it. No widget content changed.
- `cdk synth` of the pipeline stack: the prod dashboard body contains the cache widgets, the reserved-concurrency line at 3000, the GSI throttle widget and the WAF widget.
- `tsc --noEmit` clean, eslint 0 errors.

## Notes

`GetOrdersQueryCacheSize` is per execution environment and bounded by the TTL, so it reads as "distinct partitions this environment served in the last 500ms", typically a handful, not fleet-wide cardinality. Fleet-wide distinct keys come from the miss log: `filter msg = "Query cache miss" | stats count_distinct(partitionKey)`.

### DIFF
--- a/bin/stacks/dashboard-stack.ts
+++ b/bin/stacks/dashboard-stack.ts
@@ -784,6 +784,9 @@ export class DashboardStack extends cdk.NestedStack {
             x: 16,
             type: 'metric',
             properties: {
+              // WAF only counts a request under a rule's metric when the rule matches, and for a
+              // block rule that means blocked, so there is no per-rule "allowed" series. The
+              // service's own request count is the denominator instead: requests that got through.
               metrics: [
                 [
                   'AWS/WAFV2',
@@ -796,14 +799,20 @@ export class DashboardStack extends cdk.NestedStack {
                   region,
                   { label: 'ip-get-orders blocked' },
                 ],
-                ['.', 'AllowedRequests', '.', '.', '.', '.', '.', '.', { label: 'ip-get-orders allowed' }],
+                [
+                  METRIC_NAMESPACE,
+                  'GetOrdersRequest',
+                  'Service',
+                  'UniswapXService',
+                  { label: 'GetOrders requests served' },
+                ],
               ],
               view: 'timeSeries',
               stacked: false,
               region,
               stat: 'Sum',
               period: 300,
-              title: 'WAF ip-get-orders Blocked vs Allowed | 5min',
+              title: 'WAF ip-get-orders Blocked vs Requests Served | 5min',
             },
           },
           {
--- a/bin/stacks/dashboard-stack.ts
+++ b/bin/stacks/dashboard-stack.ts
@@ -271,7 +271,7 @@ export class DashboardStack extends cdk.NestedStack {
           {
             height: 6,
             width: 24,
-            y: 38,
+            y: 44,
             x: 0,
             type: 'log',
             properties: {
@@ -285,7 +285,7 @@ export class DashboardStack extends cdk.NestedStack {
           {
             height: 6,
             width: 24,
-            y: 44,
+            y: 50,
             x: 0,
             type: 'log',
             properties: {
@@ -303,7 +303,7 @@ export class DashboardStack extends cdk.NestedStack {
           {
             height: 6,
             width: 24,
-            y: 50,
+            y: 56,
             x: 0,
             type: 'log',
             properties: {
@@ -412,7 +412,7 @@ export class DashboardStack extends cdk.NestedStack {
           {
             height: 6,
             width: 12,
-            y: 32,
+            y: 38,
             x: 0,
             type: 'metric',
             properties: {
@@ -432,7 +432,7 @@ export class DashboardStack extends cdk.NestedStack {
           {
             height: 6,
             width: 12,
-            y: 32,
+            y: 38,
             x: 12,
             type: 'metric',
             properties: {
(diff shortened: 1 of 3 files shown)


bin/stacks/dashboard-stack.ts
```diff
@@ -532,6 +542,270 @@ export class DashboardStack extends cdk.NestedStack {
...
+                  'BlockedRequests',
+                  'WebACL',
+                  `${SERVICE_NAME}IPThrottling`,
+                  'Rule',
+                  'ip-get-orders',
+                  'Region',
+                  region,
+                  { label: 'ip-get-orders blocked' },
+                ],
+                ['.', 'AllowedRequests', '.', '.', '.', '.', '.', '.', { label: 'ip-get-orders allowed' }],
```
ericneil-sanc: I think the AllowedRequests series here will always be empty, because WAF only counts a request under a rule's metric when the rule matches, and for a block rule matching means blocked.

codyborn: Right, a block rule only ever reports BlockedRequests. Swapped the second series for the service's own `GetOrdersRequest` count so the chart is blocked vs. requests that got through, and added a test that the widget has no AllowedRequests series (1501eff).

bin/stacks/dashboard-stack.ts
```diff
@@ -532,6 +542,279 @@ export class DashboardStack extends cdk.NestedStack {
...
+            type: 'metric',
+            properties: {
+              metrics: [
+                [
+                  'AWS/DynamoDB',
+                  'ConsumedReadCapacityUnits',
+                  'TableName',
+                  TABLE_NAMES.Orders,
+                  'GlobalSecondaryIndexName',
+                  CHAIN_STATUS_GSI,
```
ericneil-sanc, line 721: I think this is missing the third hot GSI. CACHEABLE_INDEXES also includes the plain chainId index (chainId-createdAt-all), and a bare ?chainId= query with no status filter maps to exactly that one. Thats the shape of most of the polling traffic this section is meant to watch, so a throttle spiral there wouldnt show up on either GSI widget. Should we add it to this widget and the throttles one below?

ericneil-sanc (dismissed):

bin/stacks/dashboard-stack.ts
codyborn: Good catch, added. Both GSI widgets now plot chainId_orderStatus, orderStatus and chainId for the Orders and LimitOrders tables, matching CACHEABLE_INDEXES, and the test checks all three appear in both (7c9f318).

ericneil-sanc (approved):

PR #706, opened 2026-09-04
chore: remove Relay order support

alanhwu:
## Summary

Removes the Relay order feature end to end. Relay has never carried production traffic and the code was being built into every Lambda cold start.

## Evidence it is unused

| Check | Result |
|---|---|
| Prod `RelayOrders` table (created Apr 2024) | 0 items, 0 bytes |
| Beta `RelayOrders` table | 0 items |
| PostOrder Lambda logs mentioning Relay, last 14 days | 0 |
| CheckOrderStatus Lambda logs with `orderType: Relay`, last 14 days | 0 |
| WAF sample of 500 GET requests, last 3 hours, with `Relay` in the URI | 0 |
| Consumers on the RelayOrders DynamoDB stream | none |

## What changes

- Deleted `RelayOrderService`, `OffChainRelayOrderValidator`, `RelayOrderRepository`, `RelayOrder` model, `RelayOrderEntity`, `GetRelayOrderResponse`, and their tests and factory.
- Unwired the Relay service from the five Lambda entrypoints (`post-order`, `post-limit-order`, `get-orders`, `get-limit-orders`, `check-order-status`) and from `OrderDispatcher` and `CheckOrderStatusHandler`. Each entrypoint was constructing `RelayEventWatcher` RPC providers for mainnet, Polygon and Sepolia on every cold start.
- Removed `Relay` from the POST and GET `orderType` Joi enums and the query-param enum. Requests carrying it now get a 400 from validation instead of a 200 with an empty list.
- Deleted `EventWatcherMap`, whose only remaining use was the Relay watcher factory.
- Collapsed `OrderEntityType` and the related generics to the UniswapX entity only.
- Removed the `RelayOrderEntity` schema and enum value from `swagger.json`; updated README and CLAUDE.md.
- Removed the `RelayOrders` table construct and its capacity config from `dynamo-stack`, `bin/config.ts` and `bin/app.ts`.

## Kept on purpose

The deadline-less run cap in `CheckOrderStatusHandler` (`MAX_ORDER_TRACKING_RUNS_WITHOUT_DEADLINE`) and its metric stay. Relay was the only order type that never carried a deadline, but executions started before `deadline` was echoed into SFN state also lack one, and the cap is what bounds them. Comments updated to drop the Relay reference.

## Deploy notes

CDK's default removal policy for tables is RETAIN and deletion protection is enabled, so the deploy orphans the empty `RelayOrders` table rather than deleting it. After the deploy lands, disable deletion protection and delete the table in beta and prod by hand.

## Testing

- `tsc --noEmit`: clean
- `yarn lint`: 0 errors
- `yarn test`: 1,199 passed. `test/unit/stacks/dashboard-stack.test.ts` fails on `main` too: `bin/stacks/dashboard-stack.ts` imports `lodash`, which is not declared in `package.json`.
- `yarn test:integ` for `test/integ/handlers/check-order-status.test.ts`: passed.


test/unit/errors/UnexpectedOrderTypeError.ts
```diff
@@ -4,7 +4,7 @@ import { UnexpectedOrderTypeError } from '../../../lib/errors/UnexpectedOrderTyp
 describe('UnexpectedErrorTypeError', () => {
   it('encodes the order type into the message', () => {
     expect(() => {
-      throw new UnexpectedOrderTypeError(OrderType.Relay)
-    }).toEqual('unexpected order type: Relay')
+      throw new UnexpectedOrderTypeError(OrderType.Priority)
```
ericneil-sanc: I think this file can just be deleted, its not named *.test.ts so jest never runs it, and the assertion would fail if it did (compares a function to a string)

ericneil-sanc (dismissed):

test/unit/errors/UnexpectedOrderTypeError.ts
alanhwu: Good call. Deleted the file in 9ab5f50.

ericneil-sanc (approved):

PR #707, opened 2026-09-04
chore: remove Hybrid orders and the UniswapX V4 validator

alanhwu:
Reverts the Hybrid order support added in #636 plus the Hybrid branches that #654, #665, and #683 layered on top. Hybrid orders were only ever deployable on Unichain Sepolia (1301), prod holds zero orders on that chain, and no producer posts orderType Hybrid or a hardQuote body.

Removed:
- HybridOrder model, GetHybridOrderResponse schema, initializer client, CosigningError, and the Hybrid test factory/model tests
- V4OrderValidator plumbing (OnChainValidatorMap generic, post-order wiring, UniswapXOrderService constructor param and validate branch)
- Hybrid cases in the post/get paths, offchain validator, status tracker, reaper, Permit2Validator, analytics, entities, and field validators
- hardQuote request-body field and types
- HybridOrderEntity from swagger.json and its pin test

Chain 1301 stays in SUPPORTED_CHAINS; that list is general service infra (RPC, per-chain Step Functions, alarms, dashboards). Uniswap v4 AMM calldata handling in UniversalRouterCalldata is unrelated and untouched.


lib/util/Permit2Validator.ts
```diff
@@ -47,38 +36,36 @@ export class Permit2Validator {
...
-      }
+    if (permitValidation.isExpired) {
+      return OrderValidation.Expired
+    }
 
     // Check if order deadline has passed
     const currentTimestamp = Math.floor(Date.now() / 1000)
     if (currentTimestamp > order.info.deadline) {
-      return OrderValidation.Expired
+    return OrderValidation.Expired
```
ericneil-sanc: nit: this return is flush with the if above it, looks like the de-indent missed a line. Behavior is fine, just needs a prettier pass

ericneil-sanc (dismissed):
nit: title says feat: but this is a pure removal, chore: would match what we did for the relay removal in #706

ericneil-sanc (approved):

PR #714, opened 2026-09-22
feat(iam): cross-account roles for the backend-monorepo UniswapX service (ECO-861)

aayala-uniswap:
## Why

The UniswapX service being built in the backend monorepo needs to read, and later write, the `Orders` and `LimitOrders` tables in this account while the status-tracking logic moves before the tables do. The platform team's condition on the ask ([Slack thread](https://uniswapteam.slack.com/archives/C08JDHZGZ7B/p1789497840514909), [ECO-861](https://linear.app/uniswap/issue/ECO-861)) was that these roles be codified in the gouda account rather than created by hand. This is that.

**Draft** — @SocksNFlops please confirm the permission sets match what the comparator and status worker actually call before this leaves draft. @melodylove-uniswap for the trust shape.

## What

New nested `IamStack` under `APIStack` (`bin/stacks/iam-stack.ts`). Two roles per stage, scoped to `Orders` + `LimitOrders` only, nothing on `Nonces` or any other table:

| Role | Actions | Resources | Used when |
|---|---|---|---|
| `uniswapx-shadow-read-<stage>` | `GetItem`, `BatchGetItem`, `Query`, `DescribeTable` | both tables + `table/*/index/*` | Shadow comparator diffs the new tracker against live tables |
| | `DescribeStream`, `GetRecords`, `GetShardIterator`, `ListStreams` | both tables' streams | |
| `uniswapx-orders-write-<stage>` | `PutItem`, `UpdateItem`, `GetItem`, `Query`, `ConditionCheckItem` | both tables + indexes | After cutover. No `Delete`, no `Scan`, no batch writes. Unassumed until then; revoking it is the rollback |

Role ARNs are stack outputs (`ShadowReadRoleArn`, `OrdersWriteRoleArn`) so the backend side can pin them.

### Trust: per stage, one statement per account

The ask proposed trusting all three backend accounts from both gouda accounts. This PR tightens that to the matching tier:

- **beta** (321377678687) trusts backend **dev** 411170392337 + **staging** 413367642260
- **prod** (316116520258) trusts backend **prod** 654200013602 only
- **local** does not create the nested stack at all (an empty nested stack is rejected by CloudFormation at deploy; `APIStack` gates on `BACKEND_ACCOUNTS`)

Each trust statement names one account and is narrowed with `aws:PrincipalArn` → `arn:aws:iam::<that account>:role/uniswapx-ecsTaskRole-*`, the name `buildEcsTaskRole` gives the backend task role (`serviceShortName` is `uniswapx`). Not the wider `uniswapx-*`: that would admit every `uniswapx-`-prefixed role in the backend account and bypass the backend side's cutover flag on the write role (review, @alanhwu). The task role doesn't exist until first deploy, so trust is pinned to account + name pattern; pinning the exact ARN later is a one-line change in `BACKEND_ROLE_NAME_PATTERN`. If you need beta to be reachable from backend prod (or vice versa) for some test path, say so and I'll widen it — it's deliberate, not an oversight.

`LambdaStack` now exposes `ordersTable` / `limitOrdersTable` so `APIStack` can hand them to the new stack. `CLAUDE.md` documents the roles and the "do not widen to other tables" constraint.

## Not in this PR

- GCP Pub/Sub identity (Blair, separate).
- Anything on the backend side: the `uniswapx-*` task roles, the assume-role wiring, or the shadow comparator.
- Exact task-role ARN pinning (after the backend service's first deploy).

## Verification

- `test/unit/stacks/iam-stack.test.ts` (10 tests): exactly two roles with stable names; prod trusts only backend prod; beta trusts dev + staging with each statement conditioned on its own account's task role; the pattern is pinned to `uniswapx-ecsTaskRole-*` and every trust statement ends in it; read role has exactly the table-read + stream-read action sets on 4 table/index ARNs + 2 stream ARNs; write role has exactly the five write-path actions, one statement, 4 resources; no `Delete` / `Scan` / `BatchWrite` / table-management action anywhere; the template never mentions `Nonces`; constructing the stack for a stage with no backend accounts throws; both outputs present.
- `tsc` clean. `eslint` and `prettier` clean on the new files (`lambda-stack.ts` and `CLAUDE.md` were already non-conforming on `main`; not reformatted here to keep the diff to the change).
- `cdk synth --all`: the beta and prod nested IAM templates contain the roles, trust and policies exactly as described above (checked by hand; all four roles' conditions end in `role/uniswapx-ecsTaskRole-*`). The local `GoudaServiceStack` assembly has **no** IamStack template. No other template changes.
- Full unit suite: 40/42 suites pass. The 2 failing suites (`field-validator`, `order-validator`) fail identically on untouched `main` (Joi `valid(...SUPPORTED_CHAINS)` evaluates before the chain list initialises in the test env) and don't touch this change.

## After merge

The pipeline deploys the roles with the next beta → prod run. ARNs are then readable from the nested stack outputs, or simply:

```
arn:aws:iam::321377678687:role/uniswapx-shadow-read-beta
arn:aws:iam::321377678687:role/uniswapx-orders-write-beta
arn:aws:iam::316116520258:role/uniswapx-shadow-read-prod
arn:aws:iam::316116520258:role/uniswapx-orders-write-prod
```

Transitional: this stack is deleted with the DynamoDB migration.


aayala-uniswap:
Backend half (task role `sts:AssumeRole` on these ARNs, write role behind a cutover flag): https://github.com/Uniswap/backend/pull/14066 — stacked on #13377 until the scaffold lands.

bin/stacks/api-stack.ts
```diff
@@ -85,6 +88,12 @@ export class APIStack extends cdk.Stack {
       checkStatusFunction,
     })
 
+    new IamStack(this, `${SERVICE_NAME}IamStack`, {
```
alanhwu: Blocker: on `STAGE.LOCAL` this still creates the nested stack, but `IamStack` returns before adding anything, so its template synthesizes to `{}`. CloudFormation rejects a template with no Resources ("At least one Resources member must be defined"), so `cdk deploy GoudaServiceStack` fails and rolls back. Synth passes because CDK's synth-time validation only checks top-level templates.

Suggest gating at the call site instead:
```ts
if (BACKEND_ACCOUNTS[stage as STAGE].length > 0) {
  new IamStack(this, `${SERVICE_NAME}IamStack`, { stage: stage as STAGE, ordersTable, limitOrdersTable })
}
```
Beta and prod keep the same construct ID, so nothing gets replaced there. The early return in `IamStack` and the `?` on the role fields can then go.

bin/stacks/iam-stack.ts
```diff
@@ -0,0 +1,136 @@
...
+  [STAGE.PROD]: ['654200013602'],
+  [STAGE.LOCAL]: [],
+}
+
+/**
+ * The backend service's ECS task roles are named `uniswapx-*`. They do not exist until the
+ * service's first deploy, so trust is pinned to the account plus this name pattern rather than
+ * to exact role ARNs. Tightening to exact ARNs after first deploy is a one-line change here.
+ */
+export const BACKEND_ROLE_NAME_PATTERN = 'uniswapx-*'
```
alanhwu: Blocker: `uniswapx-*` matches every role in the backend account whose name starts with `uniswapx-`, not just the task role. Both roles get this trust. So once this deploys, any `uniswapx-*` role in 654200013602 whose own policy allows `sts:AssumeRole` can assume `uniswapx-orders-write-prod` and write prod Orders/LimitOrders, whatever `assumeGoudaOrdersWriteRole` is set to. That includes other services' roles, deploy/CI roles, or a role someone creates with `iam:CreateRole`. The flag only controls the task role's identity policy.

We already know the task role's name (`uniswapx-ecsTaskRole-<pulumi suffix>`, per Uniswap/backend#14066), and an `aws:PrincipalArn` condition doesn't need the role to exist yet. So this can be narrowed now:
```ts
export const BACKEND_ROLE_NAME_PATTERN = 'uniswapx-ecsTaskRole-*'
```
Then only the task role matches, and the backend flag becomes a real gate on the write role. The beta trust test hardcodes `role/uniswapx-*` at line 89 of the test file, so point it at the constant.

alanhwu (commented):
Two small blockers inline. The rest can wait for cutover.

bin/stacks/api-stack.ts
aayala-uniswap: Taken, in baa0839. `APIStack` now gates on `BACKEND_ACCOUNTS[stage].length > 0`, and `IamStack` throws if constructed for a stage with nothing to trust, so an empty nested template cannot be produced by accident. Role fields are non-optional again. Verified with `cdk synth --all`: the local `GoudaServiceStack` assembly has no IamStack template; beta/prod keep the same construct ID.

bin/stacks/iam-stack.ts
aayala-uniswap: Taken, in baa0839. `BACKEND_ROLE_NAME_PATTERN` is now `uniswapx-ecsTaskRole-*`, with a comment on why the wide prefix was wrong (it would have bypassed the backend cutover flag on the write role). A new test pins the constant and that every trust statement ends in it; the beta test reads the constant instead of the literal. Synth shows all four roles' conditions ending in `role/uniswapx-ecsTaskRole-*`. Backend #14066 docs updated to match.

alanhwu (approved):

